PR #10, opened 2024-05-14
adding personal details header

denis-davidov-moj:


assets/scss/local.scss
```diff
@@ -1,3 +1,38 @@
 .govuk-main-wrapper {
   min-height: 600px;
+}
+
+.offender-details {
+  margin-bottom: 25px
```
Thomas-Geraghty, line 6: Going forward, it's usually better to use the [spacing mixins provided as part of the govuk-frontend](https://design-system.service.gov.uk/styles/spacing/#applying-spacing-in-your-own-css). We can clean these up at a later date :)

server/views/pages/about-pop.njk
```diff
@@ -53,6 +54,21 @@
...
+      href: '/about-pop',
+      active: true
+    }, {
+      text: 'Goals',
+      href: '/goals'
+    }, {
+      text: 'Plan summary',
+      href: '/plan-summary'
+    }]
+  }) }}
```
Thomas-Geraghty, line 71: In future, the `text` elements should their strings stored in the locale file, but because of the flux nature of the design at the moments, we can just deal with that later

Thomas-Geraghty (approved):

PR #30, opened 2024-06-25
Rename ARNS Handover Service environment vars

Thomas-Geraghty:
- Rename the env vars to be more inline with other handover projects
- Added external URL var for when dockerized


server/authentication/auth.ts
pipwilson, line 1: where are these values populated from?

pk2453 (approved):
LGTM

server/authentication/auth.ts
Thomas-Geraghty: Apologies I can't see where your comment is pointing to, but configuration values are often stored in config.ts, with the changes for this PR existing here - https://github.com/ministryofjustice/hmpps-sentence-plan-ui/blob/10df227825ce8bca40c25e75a550d73711a8e3d0/server/config.ts#L84-L96

PR #35, opened 2024-07-01
Fix issue where server/build crashing kills concurrently

Thomas-Geraghty:
- Capture errors on build/server and print


esbuild/esbuild.config.js
```diff
@@ -61,12 +60,12 @@ function main() {
   if (args.includes('--watch')) {
     console.log('\u{1b}[1m\u{1F52D} Watching for changes...\u{1b}[0m')
     // Assets
-    chokidar.watch(['assets/**/*'], chokidarOptions).on('all', () => buildAssets(buildConfig))
+    chokidar.watch(['assets/**/*'], chokidarOptions).on('all', () => buildAssets(buildConfig).catch(console.log))
```
pipwilson: Should this (and the change on line 68) be printing to `console.error` rather than `console.log` ?

Thomas-Geraghty: Absolutely, I had the same thought earlier. Let me change those!

pipwilson (approved):
LGTM. :shipit:

PR #44, opened 2024-07-04
adding autocompete to steps

denis-davidov-moj:


assets/js/steps.js
```diff
@@ -0,0 +1,35 @@
+import accessibleAutocomplete from './accessible-autocomplete.min.js'
```
Thomas-Geraghty: I think you might be able to make these just imports from the module
`import accessibleAutocomplete from 'accessible-autocomplete'`

assets/scss/local.scss
```diff
@@ -1,3 +1,5 @@
+  @import './accessible-autocomplete.min.css';
```
Thomas-Geraghty: Same here with importing from the module
`@import "accessible-autocomplete";`

server/routes/ReferenceData/routes.ts
```diff
@@ -0,0 +1,16 @@
...
+    const { areaOfNeed } = req.params
+    res.json(referentialDataService.getSteps(areaOfNeed))
+  })
+  router.get('/reference-data/:areaOfNeed/goals', (req, res, next) => {
+    const { areaOfNeed } = req.params
+    res.json(referentialDataService.getGoals(areaOfNeed))
+  })
+  router.get('/reference-data/areasOfNeed', (req, res, next) => {
+    res.json(referentialDataService.getAreasOfNeed())
+  })
```
Thomas-Geraghty: I think you were trying to structure these a in a kind of RESTful way, this might have been a slightly better structure
```
  router.get('/reference-data/areaOfNeed/:areaOfNeed/steps', (req, res, next) => {
    const { areaOfNeed } = req.params
    res.json(referentialDataService.getSteps(areaOfNeed))
  })
  router.get('/reference-data/areaOfNeed/:areaOfNeed/goals', (req, res, next) => {
    const { areaOfNeed } = req.params
    res.json(referentialDataService.getGoals(areaOfNeed))
  })
  router.get('/reference-data/areaOfNeed', (req, res, next) => {
    res.json(referentialDataService.getAreasOfNeed())
  })
```
where you specify the resource type before the identifier, or drop the `/identifier` to get all of that type

server/routes/createGoal/routes.test.ts
Thomas-Geraghty, line 1: Considering this test has just been reduced to `expect(true)`, why keep it at all?

server/routes/steps/StepsController.ts
```diff
@@ -0,0 +1,102 @@
...
+        <h1 class="govuk-label-wrapper">
+          <label class="govuk-label govuk-label--s" for="some-one-else">
+          Enter who will do this step
+          </label>
+        </h1>
+        <input class="govuk-input govuk-!-width-one-third" id="some-one-else" name="someOneElse" type="text">
+      </div>`,
+    },
+  },
+]
```
Thomas-Geraghty: How come we're defining this in the controller rather than inside the template?

server/data/sentencePlanApiClient.ts
```diff
@@ -1,88 +1,14 @@
...
-    return SentencePlanApiClient.restClient(token).get<PlanType>({ path: `/plans/${planUuid}` })
-  }
-
-  async getPlanByOasysAssessmentPk(oasysAssessmentPk: string) {
-    logger.info(`Getting plan with OASys Assessment PK: ${oasysAssessmentPk}`)
-    const token = await this.authClient.getSystemClientToken()
-    return SentencePlanApiClient.restClient(token).get<PlanType>({ path: `/oasys/plans/${oasysAssessmentPk}` })
+  async restClient(info?: string, token?: string): Promise<RestClient> {
+    if (info) logger.info(info)
+    const realToken = token || (await this.authClient.getSystemClientToken())
```
Thomas-Geraghty: Is the `token` parameter used outside of any test? I don't see what the intended function is, unless it's just for testing purposes?

server/routes/steps/StepsController.ts
pk2453: yes this is not a controller

pk2453 (commented):
Can we have consistency on naming of ts files. Some are camel case and some are upper case and some lowercase.

assets/scss/local.scss
denis-davidov-moj: Actually, correct one is   **@import './accessible-autocomplete.min';**  otherwise it throws an error.
2024-07-05 09:52:51 [ESBuild] ✘ [ERROR] Can't find stylesheet to import.
2024-07-05 09:52:51 [ESBuild]   ╷
2024-07-05 09:52:51 [ESBuild] 1 │   @import './accessible-autocomplete'; 
2024-07-05 09:52:51 [ESBuild]   │           ^^^^^^^^^^^^^^^^^^^^^^^^^^^

assets/js/steps.js
denis-davidov-moj: If you want to remove extension correct ones is import accessibleAutocomplete from './accessible-autocomplete.min'
2024-07-05 09:53:21 [ESBuild] ✘ [ERROR] Could not resolve "./accessible-autocomplete"
2024-07-05 09:53:21 [ESBuild] 
2024-07-05 09:53:21 [ESBuild]     assets/js/steps.js:1:35:
2024-07-05 09:53:21 [ESBuild]       1 │ import accessibleAutocomplete from './accessible-autocomplete'

server/data/sentencePlanApiClient.ts
denis-davidov-moj: Added token parameter for backward compatibility since legacy code had it. Removing it!

assets/js/steps.js
pipwilson: I don't think this is right? Shouldn't this be a reference to the package rather than the static file in the directory? i.e. the line should be as Thomas puts it - with no `./` at the start and no `.min.js` at the end - in fact do we need the .js and .css files in the repo at all? Shouldn't they get bundled in via esbuild?

denis-davidov-moj: Css has thrown an error and js did't bundle autocomplete to final file.
```
2024-07-05 12:11:25 [ESBuild] ✘ [ERROR] Can't find stylesheet to import.
2024-07-05 12:11:25 [ESBuild]   ╷
2024-07-05 12:11:25 [ESBuild] 1 │   @import 'accessible-autocomplete'; 
2024-07-05 12:11:25 [ESBuild]   │           ^^^^^^^^^^^^^^^^^^^^^^^^^
2024-07-05 12:11:25 [ESBuild]   ╵
2024-07-05 12:11:25 [ESBuild]   assets/scss/local.scss 1:11        @import
2024-07-05 12:11:25 [ESBuild]   assets/scss/application.scss 13:9  root stylesheet [plugin sass-plugin]
```

denis-davidov-moj: unfortunately this package was developed originally as react package and later forked manually to GDS backend.
Ideally we should create nunjuck macro which will do all the mentioned things under the hood.

pk2453 (approved):

PR #46, opened 2024-07-05
Fix the URL for retrieving plans so the app works cleanly from OAStub

pipwilson:


Thomas-Geraghty:
Didn't we already do this in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/42 ? Or did it get merged out on accident during https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/44 ?

Thomas-Geraghty (approved):

PR #50, opened 2024-07-08
Use containerised setup in CircleCI

Thomas-Geraghty:
- Spin up a test environment using `make test-up` that includes all the pods necessary for testing our service e2e
  - probably some of these don't need spinning up (forward-proxy)
- Run make `test e2e-ci` against span up test environment
- Added run `make e2e` for running e2e/integration tests on local dev environment
- Disabled some unused healthcheck tests, and manageUsersApi health check
- Added some extra health checks (partly to test this works 😄)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,6 +3,7 @@ version: 2.1
 orbs:
   hmpps: ministryofjustice/hmpps@8
   slack: circleci/slack@4.12.5
+  mem: circleci/rememborb@0.0.2
 
 parameters:
   alerts-slack-channel:
@@ -21,7 +22,7 @@ parameters:
     default: 20.11-browsers
 
 jobs:
-  build:
+  lint:
     executor:
       name: hmpps/node
       tag: << pipeline.parameters.node-version >>
@@ -41,18 +42,8 @@ jobs:
             - node_modules
             - ~/.cache
       - run:
-          command: |
-            npm run build
-      - run: # Run linter after build because the integration test code depend on compiled typescript...
           name: Linter check
           command: npm run lint
-      - persist_to_workspace:
-          root: .
-          paths:
-            - node_modules
-            - build
-            - dist
-            - assets/stylesheets
 
   unit_test:
     executor:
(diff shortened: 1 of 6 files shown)


makefile
```diff
@@ -38,6 +38,25 @@ test: ## Runs the unit test suite.
 	@make install-node-modules
 	docker compose ${DEV_COMPOSE_FILES} run --rm --no-deps ui npm run test
 
+BASE_URL ?= "http://localhost:3000"
```
pipwilson, line 41: Is there a reason you don't use `CYPRESS_BASE_URL` here?

.circleci/config.yml
```diff
@@ -70,61 +61,67 @@ jobs:
...
+    working_directory: ~/app
     steps:
       - checkout
       - attach_workspace:
           at: ~/app
       - run:
-          name: Install missing OS dependency
-          command: sudo apt-get install libxss1
+          name: Extract saved container image
+          command: docker load --input docker_cache/build_image.tar
```
pipwilson, line 75: I don't understand this line - how do you know you're getting the right image?

makefile
Thomas-Geraghty: As in, why don't we set the `CYPRESS_BASE_URL` rather than specify `npx cypress open -c baseUrl=$(BASE_URL) ...` ?

.circleci/config.yml
Thomas-Geraghty: https://github.com/ministryofjustice/hmpps-circleci-orb/blob/7d31d821ecb9df1dd4aec768491e637f4d93e718/src/jobs/build_docker.yml#L101-L108

Using our circleCI orb, when using the `persist` option, it saves the built container to this location

pipwilson: ah awesome.

makefile
pipwilson: Yes. If you have CYPRESS_BASE_URL set locally it will override the baseUrl parameter.

Thomas-Geraghty: I guess it just looks less confusing, it is just copied from the SAN project so I'm not fussed if you want to change it. Not sure if it has a knock on effect on the any further commands in the makefile though

pipwilson: OK I can live with it - might need to bear it in mind if anyone has any problems.

pipwilson (approved):
LGTM

PR #56, opened 2024-07-11
SP2-270 - Minor fixes to create-steps

pk2453:


Thomas-Geraghty (dismissed):
I think there's still a few issues with headings being too big. Also it looks like you've managed to set the 3 character popup enforcement on the autocomplete component, but I'm wondering how you've achieved this? I don't see it added on in the JS

Thomas-Geraghty:
Pushed fixes now for my previous comment, https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/56/commits/e9e907ffdbfe5b18d4fb14b4a6969c3568ed0cdd

pk2453:
> I think there's still a few issues with headings being too big. Also it looks like you've managed to set the 3 character popup enforcement on the autocomplete component, but I'm wondering how you've achieved this? I don't see it added on in the JS

assets/js/accessible-autocomplete.min.js its updated here

pk2453 (commented):
LGTM your changes @Thomas-Geraghty

pipwilson:
> > I think there's still a few issues with headings being too big. Also it looks like you've managed to set the 3 character popup enforcement on the autocomplete component, but I'm wondering how you've achieved this? I don't see it added on in the JS
> 
> assets/js/accessible-autocomplete.min.js its updated here

I don't think we should be modifying this file - we should be using the minLength parameter (https://github.com/alphagov/accessible-autocomplete?tab=readme-ov-file#minlength-default-0) when creating the component in steps.js

pipwilson (approved):
I think there are still lots of problems with the overall output of this story and we'll need to open a new story (maybe in tech debt?) to fix both the problems here and update the design to the signed-off version, but hopefully along with the discussions with PO this should be good enough to get through our current review.

PR #61, opened 2024-07-17
SP2-476 - modify create goal page

pk2453:
All validation covered and values are preserved on error except auto complete input. Which will be looked separately


integration_tests/e2e/create-goal.cy.ts
```diff
@@ -19,4 +24,32 @@ describe('Create a new Goal', () => {
...
+    cy.url().should('include', '/create-goal/accommodation')
+    cy.contains('#other-area-of-need-radio-error', 'You need to select an area of need')
+  })
+
+  it('Creates a new goal without steps', () => {
+    cy.url().should('include', '/about-pop')
+    cy.get("a[href='/create-goal/accommodation']").click()
+    cy.url().should('include', '/create-goal/accommodation')
+    // eslint-disable-next-line cypress/unsafe-to-chain-command
+    cy.get('#goal-name').type('acc').trigger('keydown', { keyCode: 17 })
```
pipwilson: what does this line do?

assets/js/steps.js
Thomas-Geraghty, line 1: Sidenote and not to be fixed in this PR - I think we should probably rename this from `steps.js` to something else... maybe `autocomplete.js`? It covers more than steps after all.

cypress/downloads/downloads.html
Thomas-Geraghty, line 1: I think we need to remove this, and add a .gitignore override for the `cypress` folder

integration_tests/e2e/create-goal.cy.ts
```diff
@@ -4,13 +4,18 @@ describe('Create a new Goal', () => {
...
   })
 
-  it('Creates a new goal', () => {
+  it('Creates a new goal with steps', () => {
     cy.url().should('include', '/about-pop')
     cy.get("a[href='/create-goal/accommodation']").click()
     cy.url().should('include', '/create-goal/accommodation')
-    cy.get('.govuk-radios').first().get('input[value="Accommodation Goal 3"]').click()
+    // eslint-disable-next-line cypress/unsafe-to-chain-command
+    cy.get('#goal-name').type('acc').trigger('keydown', { keyCode: 17 })
```
Thomas-Geraghty: If this is for triggering the autocomplete, might it be worth doing a check after this to make sure the input field has the expected value selected?

server/data/hmppsAuditClient.test.ts
```diff
@@ -57,7 +57,7 @@ describe('hmppsAuditClient', () => {
       expect(actualMessageBody).toEqual(expectedSqsMessageBody)
 
       const eventTime = Date.parse(actualMessageBody.when)
-      expect(Date.now() - eventTime).toBeLessThan(20)
+      expect(Date.now() - eventTime).toBeLessThan(40)
```
Thomas-Geraghty:
```suggestion
      expect(eventTime).not.toBeNaN()
      expect(eventTime).toBeGreaterThan(-1)
      expect(Date.now() - eventTime).toBeLessThan(1000)
```
This is the fix that went into the typescript template project recently, if we just want to use this instead?

server/views/partials/layout.njk
```diff
@@ -15,4 +15,5 @@
 
 {% block bodyEnd %}
   <script type="module" src="/assets/js/app.js"></script>
+  <script src="/assets/javascript/datepicker.js"></script>
```
Thomas-Geraghty: Is this being added through `hmpps-court-cases-release-dates-design`? We should just be bundling this into our app.js

integration_tests/e2e/create-goal.cy.ts
pipwilson: is 17 the down arrow? if so can't we use `type({downArrow})` ?

pipwilson:
Should this PR be correcting the use of "Joan" throughout the create goal page to the person's name?

assets/js/steps.js
pk2453: renamed as this was confusing to me as well.

cypress/downloads/downloads.html
pk2453: > I think we need to remove this, and add a .gitignore override for the `cypress` folder

done

integration_tests/e2e/create-goal.cy.ts
pk2453: > If this is for triggering the autocomplete, might it be worth doing a check after this to make sure the input field has the expected value selected?

tried to check content but seems not working though UI showing text selected. Will figure out this later.

server/data/hmppsAuditClient.test.ts
pk2453: Updated

server/views/partials/layout.njk
pk2453: > Is this being added through `hmpps-court-cases-release-dates-design`? We should just be bundling this into our app.js

Tried this but seems not worked. Should it be app.ts or index.ts

integration_tests/e2e/create-goal.cy.ts
pk2453: > is 17 the down arrow? if so can't we use `type({downArrow})` ?

replaced

integration_tests/e2e/create-goal.cy.ts
pk2453: > what does this line do?

type acc on auto complete field, press key down and then enter to select 1st value

pk2453:
> Should this PR be correcting the use of "Joan" throughout the create goal page to the person's name?

as far as I know

pipwilson:
> > Should this PR be correcting the use of "Joan" throughout the create goal page to the person's name?
> 
> as far as I know

OK, so I don't think it does. If I run it up locally I get different names on basically every page:

Homepage: "About Christopher"
Create goal: "Create a goal with Joan"
Add steps: "Add steps to Ernser's goal" (step actor lists “Christopher”)
Summary: "Joan's plan"

server/views/partials/layout.njk
Thomas-Geraghty: Looking at the project, they don't export the `date-picker.js` file. Every other project that seems to be using the date-picker seems to just be adding its assets to their app manually and bundling them/serving them. I think this would be a better option for us too, at least until the date-picker becomes part of the official MOJ framework.

pk2453:
> > > Should this PR be correcting the use of "Joan" throughout the create goal page to the person's name?
> > 
> > 
> > as far as I know
> 
> OK, so I don't think it does. If I run it up locally I get different names on basically every page:
> 
> Homepage: "About Christopher" Create goal: "Create a goal with Joan" Add steps: "Add steps to Ernser's goal" (step actor lists “Christopher”) Summary: "Joan's plan"

Updated

Thomas-Geraghty:
Looking at the designs, we're missing
image
image

and this header is too big
image
I think the sidebar is too wide also

pk2453:
> Looking at the designs, we're missing image image
> 
> and this header is too big image I think the sidebar is too wide also

header updated and for navigation I am just following one third and two third

server/views/pages/create-goal.njk
```diff
@@ -70,110 +63,161 @@
...
 
                     <form id="create-goal-form" method="POST">
                         <input type="hidden" name="_csrf" value="{{ csrfToken }}">
-                        <input type="hidden" name="area-of-need" value="{{ data.referenceData.name }}">
+                        <input type="hidden" id="_areaOfNeed" value="{{ data.areaOfNeed }}">
+                        <input type="hidden" name="area-of-need" value="{{ data.displayAreaOfNeed }}">
+                        <input type="hidden" id="page" value="goals">
+                        <div class="govuk-form-group">
+                            <h1 class="govuk-label-wrapper">
+                                <label class='govuk-label--m' for="step-name">What goal should {{ data.popData.firstName }} try to achieve?</label>
```
pipwilson: Should this be `givenName` ? When I run it up locally all usages of firstName render as empty or undefined.

server/views/pages/create-goal.njk
```diff
@@ -70,110 +63,161 @@
...
+                              ]
                         }) }}
 
                         {{ govukRadios({
-                            name: "date-selection-radio",
+                            name: "start-working-goal-radio",
                             fieldset: {
                                 legend: {
-                                    text: locale.dateSelection.label | replace("{POP_NAME}", data.popData.firstName),
+                                    text: locale.startWorking.label | replace("{POP_NAME}", data.popData.firstName),
```
pipwilson: givenName as above

server/views/partials/layout.njk
pk2453: ok this is done

pipwilson:
I know our error handling is sketchy so unsure if this should be addressed here, but if I try and submit the goal when I say there is a target date but don't select an option then it correctly throws a validation error with "You need to select a timeframe" but it wipes the title of the goal I've entered into the autocomplete field (both when I select from the autocomplete list or type something by hand).

pk2453:
> I know our error handling is sketchy so unsure if this should be addressed here, but if I try and submit the goal when I say there is a target date but don't select an option then it correctly throws a validation error with "You need to select a timeframe" but it wipes the title of the goal I've entered into the autocomplete field (both when I select from the autocomplete list or type something by hand).

Well the scope of ticket was happy scenario, all error validation and data preserve I did on top of it. As mentioned in description auto-complete filed is not included in this as its custom one and will look into that later

Thomas-Geraghty:
>I am just following one third and two third

it looks like the design is 1 quarter, 3 quarter - it'd be best to try match the design 😃

pk2453:
> > I am just following one third and two third
> 
> it looks like the design is 1 quarter, 3 quarter - it'd be best to try match the design 😃

prototype is also same one third and  two third

server/views/pages/create-goal.njk
pk2453: updated

assets/js/index.js
```diff
@@ -1,6 +1,7 @@
 import * as govukFrontend from 'govuk-frontend'
 import * as mojFrontend from '@ministryofjustice/frontend'
-import './steps.js'
+import './autocomplete.js'
+import './datepicker.js'
```
pipwilson, line 4: instead of making our own copy I think we could do `import 'hmpps-court-cases-release-dates-design/hmpps/assets/javascript/datepicker.js'` - can you try?

integration_tests/e2e/create-goal.cy.ts
```diff
@@ -19,4 +24,32 @@ describe('Create a new Goal', () => {
     cy.get('button').contains('Save and continue').click()
     cy.url().should('include', '/summary')
   })
+
+  it('Creates a new goal with errors', () => {
+    cy.url().should('include', '/about-pop')
+    cy.get("a[href='/create-goal/accommodation']").click()
+    cy.url().should('include', '/create-goal/accommodation')
+    // eslint-disable-next-line cypress/unsafe-to-chain-command
+    cy.get('#goal-name').type('acc').type('{downArrow},{enter}')
```
pipwilson: now this isn't using keydown you can merge the two `type` command to `.type('acc{downArrow},{enter}')` and remove the eslint config - same for the two other instances of this below

assets/js/index.js
pk2453: It was like this only but its suggestion from Tom
"Looking at the project, they don't export the date-picker.js file. Every other project that seems to be using the date-picker seems to just be adding its assets to their app manually and bundling them/serving them. I think this would be a better option for us too, at least until the date-picker becomes part of the official MOJ framework"

integration_tests/e2e/create-goal.cy.ts
pk2453: updated but lint config still required as its still complaining

assets/js/index.js
pipwilson: Hm OK maybe let's pick it up with @Thomas-Geraghty when he's back. I'll resolve for now though.

integration_tests/e2e/create-goal.cy.ts
pipwilson: Fixed in [3ebd114](https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/61/commits/3ebd1144be53791fc316757045e82b901a66b633)

pipwilson (approved):
🚀

PR #62, opened 2024-07-19
SP2-279 implement error summary

pk2453:


alexw-moj:
Not working as I'd expect. Selecting a goal (without selecting any combos), and hitting `Save without steps` will return a page with only two errors on, but the input is empty. Immediately hitting `Save without steps` again will now show three errors. According to https://design-system.service.gov.uk/patterns/validation/#how-to-tell-the-user-about-validation-errors ,  we should `show them the page again, with the form fields as the user filled them in`.

pk2453:
> Not working as I'd expect. Selecting a goal (without selecting any combos), and hitting `Save without steps` will return a page with only two errors on, but the input is empty. Immediately hitting `Save without steps` again will now show three errors. According to https://design-system.service.gov.uk/patterns/validation/#how-to-tell-the-user-about-validation-errors , we should `show them the page again, with the form fields as the user filled them in`.

There is no way you can prep-oulate autocomplete field so this behaviour is expected.

pk2453:
> Not working as I'd expect. Selecting a goal (without selecting any combos), and hitting `Save without steps` will return a page with only two errors on, but the input is empty. Immediately hitting `Save without steps` again will now show three errors. According to https://design-system.service.gov.uk/patterns/validation/#how-to-tell-the-user-about-validation-errors , we should `show them the page again, with the form fields as the user filled them in`.

This was a known issue but finally sorted

alexw-moj (approved):
Validation works as expected.

PR #73, opened 2024-07-31
Fix goal, steps and plan types

Thomas-Geraghty:
- We really need to sort out types out
- Fixed a bunch of types to match the API
- Fixed and updated test data (I hate this test data thing now!)

**Note: I think NewStep and Step could be made into one file, same with NewGoal and Goal - but it would end up changing a bunch of other file imports so worth doing as a separate PR**


server/@types/GoalType.ts
```diff
@@ -1,14 +1,17 @@
...
+  areaOfNeed: AreaOfNeed
+  relatedAreasOfNeed: AreaOfNeed[]
   creationDate: string
   completedDate?: string
   targetDate: string
-  isAgreed?: boolean
-  agreementNote: string
   goalOrder?: number
   moveUpURL?: string
   moveDownURL?: string
```
Thomas-Geraghty, line 11: I don't get why these are part of the `Goal` object? This data is only used for one view, it shouldn't have leaked into the common object.

server/@types/GoalType.ts
```diff
@@ -1,14 +1,17 @@
...
 }
+
+type AreaOfNeed = {
+  uuid: string
```
Thomas-Geraghty, line 15: Also we shouldn't be returning the UUID from the area of need to the view/frontend.

server/@types/NewStepType.ts
```diff
@@ -1,5 +1,10 @@
...
   description?: string
-  actor: string[]
+  actor: Actor[]
   status?: string
 }
+
+type Actor = {
+  actor: string
+  actorOptionId: number
+}
```
Thomas-Geraghty, line 10: If we combined NewStepType and StepType, this type could be shared across both 😄 - one for the future

pk2453 (approved):

server/@types/GoalType.ts
pipwilson: Moved into https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/76

PR #76, opened 2024-08-01
SP2-583 add move up down functionality

pk2453:


pipwilson:
@pk2453 can you add some cypress tests for this functionality please?

server/utils/utils.ts
```diff
@@ -81,5 +81,5 @@ export function moveGoal(goals: Array<any>, uuid: string, operation: string) {
   ;[orderedGoals[index], orderedGoals[targetIndex]] = [orderedGoals[targetIndex], orderedGoals[index]]
   orderedGoals[index].goalOrder = index
   orderedGoals[targetIndex].goalOrder = targetIndex
```
pipwilson, line 83: So I _think_ this is OK but because Goals can be deleted and we don't change the goal order of the other goals in the plan when that happens, the initial goalOrder values and index position in orderedGoals won't necessarily be the same. Are there any edge cases we need to be aware of, or is it fine as-is?

server/routes/plan-summary/PlanSummaryController.ts
```diff
@@ -42,10 +42,11 @@ export default class PlanSummaryController {
     try {
       const planUuid = req.services.sessionService.getPlanUUID()
       const goals = await this.goalService.getGoals(planUuid)
```
pipwilson, line 44: if this method was accessed via POST couldn't the correct list of goals be sent in the body, saving us this lookup and the selection from the map on line 46?

server/@types/GoalType.ts
```diff
@@ -7,8 +7,6 @@ export type Goal = {
   completedDate?: string
   targetDate: string
   goalOrder?: number
```
pipwilson: this is now non-nullable in the backend so can remove `?`

server/services/sentence-plan/goalService.ts
```diff
@@ -26,8 +27,8 @@ export default class GoalService {
     return restClient.get<Goals>({ path: `/plans/${parentPlanUuid}/goals` })
   }
 
-  async changeGoalOrder(goals: Goal[]) {
+  async changeGoalOrder(goalOrders: GoalOrder[]) {
     const restClient = await this.sentencePlanApiClient.restClient('Reordering goals')
-    return restClient.post<Goal[]>({ path: `/goals/order`, data: goals })
+    return restClient.post({ path: `/goals/order`, data: goalOrders })
```
pipwilson, line 32: At the backend we accept an unlimited number of GoalOrder objects in the list - is that OK, or will it always be two, and if so should there be some form of validation (either here or in the backend)?

pk2453: as soon as you click move up or down so it will hit the backend and update, so logically only 2 will go in one call, so dont think any change required on backend too even if it supports multiple

server/utils/utils.ts
pk2453: Dont think any issue with this as it only reverse the 2 goal order. May be on remove goal, we could reorder all goals again but this will not be impacted.

server/@types/GoalType.ts
pk2453: updated

server/routes/plan-summary/PlanSummaryController.ts
pk2453: No, because its used from tabs as well which is GET so dont want to handle things differently at 2 different places for same thing and same can be reused  while removing goal

pipwilson (approved):
LGTM

PR #77, opened 2024-08-02
Redirects back to correct tab when a goal is removed

pipwilson:


pk2453 (commented):
When I cancel, it takes to current tab when I was changing from future tab. So may need to handle similar thing for cancel button

neiltait (dismissed):

pipwilson:
> When I cancel, it takes to current tab when I was changing from future tab. So may need to handle similar thing for cancel button

@pk2453 Fixed in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/77/commits/4265561def70a1570c71831925ec641432d737dc

pk2453 (approved):

PR #82, opened 2024-08-07
Add error-summary macro

Thomas-Geraghty:
- Added error-summary macro which can be passed the errors object generated from the validation middleware
- Cleaned up confusing getFormattedError nunjucks global function
- Added href anchor tag link to getFormattedError output


pipwilson (approved):
😍

PR #83, opened 2024-08-07
Cleanup autocomplete

Thomas-Geraghty:
- Cleaned up autocomplete input and JS for goals
- Cleaned up autocomplete input and JS for steps
- Separated these two out into separate files, as, although currently they are the same, for supporting multiple autocomplete boxes on one page (for the new add-steps), the code will end up being different enough that this change would have been necessary anyway
- Cleaned up no longer needed code in the goals controller


pk2453 (approved):
LGTM

PR #86, opened 2024-08-09
SP2-513 cypress page objects

alexw-moj:
Adds additional Cypress tests, reworking to use Page Objects, and also fixes a few bugs that would prevent the tests running.


cypress.config.ts
```diff
@@ -17,6 +17,16 @@ export default defineConfig({
       on('task', {
         reset: resetStubs,
         ...auth,
+        log(message) {
+          // eslint-disable-next-line no-console
+          console.log(message)
```
pk2453: Can we not use logger instead of console.log?

assets/scss/components/summary-card.scss
```diff
@@ -3,4 +3,9 @@
   @include govuk-media-query($from: "tablet") {
     padding-left: govuk-spacing(2);
   }
+}
+
+.govuk-summary-card__actions {
```
pipwilson, line 8: is this related to the cypress tests??

server/routes/createGoal/CreateGoalController.ts
```diff
@@ -24,13 +24,18 @@ export default class CreateGoalController {
...
+    try {
+      const { uuid } = await this.goalService.saveGoal(processedData, planUuid)
+      req.services.formStorageService.saveFormData('currentGoal', {
+        processed: null,
+        raw: { uuid },
+      })
+      const redirectUrl: string =
+        req.body.action === 'addStep' ? URLs.CREATE_STEP : `${URLs.PLAN_SUMMARY}?status=success`
+      return res.redirect(redirectUrl)
+    } catch (e) {
```
pipwilson, line 36: what's the Error this catches? In fact, is it an Error or something else?

docker/docker-compose.yml
```diff
@@ -48,7 +48,7 @@ services:
     ports:
       - "7070:7070"
     healthcheck:
-      test: ["CMD", "curl", "-f", "http://arns-handover:7070/health/ping"]
+      test: ["CMD", "wget", "-q", "-O", "-", "http://localhost:7070/health/ping"]
```
pipwilson: Why the switch to wget here? Using `-O -` is the default curl behaviour so I'm a bit confused by this.

Also why the switch to localhost here?

integration_tests/e2e/plan-summary.cy.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  it('Creates three new goals, and moves the middle goal up', () => {
+    createGoalPage.createCompleteGoal(0)
+    createGoalPage.createCompleteGoal(1)
+    createGoalPage.createCompleteGoal(2)
+    planSummary.clickUpOnSummaryCard(1)
+
+    planSummary
+      .getSummaryCard(0)
+      .should('contain', 'Accommodation Goal 2')
```
pipwilson: I don't like that we're testing the title based on the autocomplete options here - we know these will change. Can you set the goal title with a parameter in "createCompleteGoal" and then test for that value here? I think it would also make the test clearer. Ditto the other tests in this file.

integration_tests/e2e/plan-summary.cy.ts
```diff
@@ -0,0 +1,70 @@
...
+    createGoalPage.addGoalAutoCompletionText(
+      'This is an example of an extremely long goal title, a goal title that some may say is too long. This goal is far longer.',
+    )
+    createGoalPage.selectOtherAreasOfNeedRadio('no')
+    createGoalPage.selectStartWorkingRadio('yes')
+    createGoalPage.selectAchievementDate('In 6 months')
+    createGoalPage.clickButton('Save without steps')
+
+    planSummary.getSummaryCard(0).within(() => {
+      cy.get('.govuk-summary-card__action').first().invoke('height').should('equal', 25)
```
pipwilson: Is the intent here that we have a suite of layout tests powered by Cypress to test overflows?

integration_tests/pages/create-goal.ts
```diff
@@ -0,0 +1,57 @@
...
+    this.createGoal('accommodation')
+    this.selectGoalAutocompleteOption('acc', value)
+    this.selectOtherAreasOfNeedRadio('no')
+    this.selectStartWorkingRadio('yes')
+    this.selectAchievementDate('In 6 months')
+    this.clickButton('Save without steps')
+  }
+
+  selectFutureGoalsSubNavigation = () => {
+    cy.get('.moj-sub-navigation__list > li:nth-child(2)').click()
```
pipwilson: I think we'd be better adding some addressable IDs to the elements on the relevant page and getting it directly. e.g currentGoals and futureGoals or something

assets/scss/components/summary-card.scss
alexw-moj: This is required to conform to the design in Figma, which is tested for in the 'height' test below.

server/routes/createGoal/CreateGoalController.ts
alexw-moj: A `SanitisedError`. The call to `saveGoal()` will fail (as the API will reject it as being too long for the column definition) if a goal title of greater than 128 chars is sent. The app falls over if this happens, as express 4 will not handle the async error.

docker/docker-compose.yml
alexw-moj: `curl` has been added additionally to the docker image. This is not necessary, as we can use `wget` without installing anything. I will revert the change to `localhost`

integration_tests/e2e/plan-summary.cy.ts
alexw-moj: The Figma design specifies that the actions should stay in line with long goal titles. I wanted to capture that as part of a test. Do you have any suggestions of a better way to test that behaviour?

integration_tests/pages/create-goal.ts
alexw-moj: Those id would only be used for targeting by the test suite, and not the app. Not a fan of adding ids unused by the app itself.

cypress.config.ts
alexw-moj: This is actually intentional, so you can see data in the console log on a test step.

integration_tests/e2e/plan-summary.cy.ts
alexw-moj: I will address this.

assets/scss/components/summary-card.scss
pipwilson: Generally speaking I think it would have been better to have this in a separate PR.

integration_tests/pages/create-goal.ts
pipwilson: My understanding is that it's the recommended approach by most QA teams using browser automation tools, since it provides a layer of isolation between the specific page layout (e.g. we add a new "overdue" tab between "current" and "future") and element that we want to interact with. In fact most recommendations I've seen have been to add a `data-qa` or `data-testautomation` attribute, but I'm really not a fan of that unless IDs are auto generated by a tool that is creating the markup for you on the fly.

My view is that an ID only used for testing is OK compared to a potentially fragile test, but I'll leave it up to you.

cypress.config.ts
pipwilson: Isn't this the default behaviour?

server/routes/createGoal/CreateGoalController.ts
pipwilson: awesome.

integration_tests/e2e/plan-summary.cy.ts
pipwilson: 1. I think the test has a typo - the selector should be against `govuk-summary-card__actions` to check that the height of the whole list is 25, not just a single item since even on the current `main` each item always has a height of 25 even if the list wraps.
2. We should check that this is the intended behaviour for tablet sizes as well as desktop since it means the goal title wraps to make way for the actions (but make the change in a future PR if necessary)
3. I think this is a case for visual regression testing against a baseline that we set per-test per-container-size rather than in raw cypress like this - I think there was some discussion of this the other week and we said we'd open a spike in tech debt?

docker/docker-compose.yml
pipwilson: Should the same change apply to the other two uses of curl in this file?

I think this should have been in a different PR, but you could use the PR description field in future to list what the changes this PR covers are.

cypress.config.ts
Thomas-Geraghty: Does this modify `cy.log()` somehow? I believe that's the intended way of logging to the console in Cypress

alexw-moj: It does not modify `cy.log()`, it is just for task() calls, e.g. https://docs.cypress.io/api/commands/task#Arguments. I have removed `log` as it was unused.

integration_tests/pages/create-goal.ts
alexw-moj: I'm not a fan of using more id's than necessary, as they can be targeted using anchors in the browser. https://docs.cypress.io/guides/references/best-practices#Selecting-Elements does suggest doing it with additional attributes for testing.

Thomas-Geraghty: I think the safe way to write this is like

`cy.get('.moj-sub-navigation__list > li').contains('Future goals').click();`

No need for additional IDs, but continues working if new tabs are added or shift around.

integration_tests/pages/plan-summary.ts
```diff
@@ -0,0 +1,15 @@
...
+  clickUpOnSummaryCard = (value: number) => {
+    this.getSummaryCard(value).contains('.govuk-button', 'Move goal up').click()
+  }
+
+  clickDownOnSummaryCard = (value: number) => {
+    this.getSummaryCard(value).contains('.govuk-button', 'Move goal down').click()
+  }
+}
+
+export default PlanSummary
```
Thomas-Geraghty: Any reason you've done the `export default` declaration separately to the class? It's much neater to just write it like 
```
export default class PlanSummary {
....
]
```

integration_tests/pages/create-goal.ts
alexw-moj: Amended.

integration_tests/pages/plan-summary.ts
alexw-moj: Amended.

integration_tests/e2e/plan-summary.cy.ts
```diff
@@ -0,0 +1,80 @@
...
+      cy.wrap(planDetails).as('planDetails')
+      cy.openSentencePlan(planDetails.oasysAssessmentPk)
+      cy.visit('/plan-summary?source=nav')
+    })
+  })
+
+  it('Creates three new goals, and moves the middle goal up', () => {
+    createGoalPage.createCompleteGoal(1)
+    createGoalPage.createCompleteGoal(2)
+    createGoalPage.createCompleteGoal(3)
```
Thomas-Geraghty: Isn't this re-testing the create goal flow a bunch? Why are we doing this if we have tests specific for create-goal? I think it'd be more appropriate to setup the test data through the API, similar to `remove-goal`,

```
      cy.get<{ plan: PlanType }>('@plan').then(({ plan }) => {
        cy.addGoalToPlan(plan.uuid, DataGenerator.generateGoal())
        cy.addGoalToPlan(plan.uuid, DataGenerator.generateGoal())
        cy.addGoalToPlan(plan.uuid, DataGenerator.generateGoal())
      })
```

alexw-moj: Amended to use this approach.

integration_tests/e2e/plan-summary.cy.ts
alexw-moj: 1. Changed the test to check the top of the first and last action elements to make sure there is no wrap.
2. https://medium.com/@pratik_shrestha/responsive-design-testing-with-cypress-a-comprehensive-guide-fc57f33134c5 shows use of viewport for after we have checked the requirement for other screen sizes.
3. https://docs.cypress.io/plugins/#visual-testing, but can remove this test if you prefer?

pipwilson: I discussed this with Tom, we previously agreed we wouldn't add visual regression testing, but I'd forgotten that - sorry. Given the small number of custom components we agreed that it wasn't worth the overhead of maintaining layout-based tests in code and so I will remove this test and we've all learned a lot about cypress in the meantime :)

Thomas-Geraghty (approved):

PR #90, opened 2024-08-09
Fix issue with cypress race condition with autocomplete fields

Thomas-Geraghty:
Fix for small bug introduced in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/83 that would cause Cypress to get confused as to which field to select from the autocomplete, as both would exist for a small amount of time.

### DIFF
--- a/assets/js/autocomplete/goal-autocomplete.js
+++ b/assets/js/autocomplete/goal-autocomplete.js
@@ -19,6 +19,8 @@ async function initializeGoalInputAutocomplete() {
     const source = await getGoalOptionsByAreaOfNeed(areaOfNeed);
     const wrapperElement = document.querySelector(`.${GOAL_AUTOCOMPLETE_WRAPPER_CLASS}`)
     const inputElement = document.querySelector(`#${GOAL_AUTOCOMPLETE_INPUT_ID}`)
+    const defaultValue = inputElement?.value ?? ''
+    inputElement.remove()
 
     accessibleAutocomplete({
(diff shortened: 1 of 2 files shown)


assets/js/autocomplete/goal-autocomplete.js
```diff
@@ -19,15 +19,15 @@ async function initializeGoalInputAutocomplete() {
     const source = await getGoalOptionsByAreaOfNeed(areaOfNeed);
     const wrapperElement = document.querySelector(`.${GOAL_AUTOCOMPLETE_WRAPPER_CLASS}`)
     const inputElement = document.querySelector(`#${GOAL_AUTOCOMPLETE_INPUT_ID}`)
+    const defaultValue = inputElement?.value ?? ''
```
pipwilson, line 22: will this trigger any negative behaviour like the autocomplete popping up on page load if e.g. my goal title in the Accommodation AoN is "Accommodation" ?

Thomas-Geraghty: This just sets loads a text value into the autocomplete box when it's loaded. It grabs the existing input element's value, or blank if non/undefined, and puts that into the autocomplete input box. The autocomplete box only expands when the element is focused so it shouldn't cause it to automatically expand on page load.

pipwilson: Great.

pipwilson (approved):
🚢

PR #100, opened 2024-08-15
SP2-607 Create the Agree Plan controller routing

pipwilson:


pk2453 (commented):
LGTM... should we also start adding unit tests for controller as part this rather than separately?

pipwilson:
I did look at that, but because there isn't a view yet or any action to take I didn't add it. I will add a skeleton controller test file.

pk2453 (approved):

PR #102, opened 2024-08-16
SP2-608 Agree Plan page

pipwilson:
Adds the web page and its validation. Backend endpoint does not yet exist so some tests are left as TODO.


pk2453:
Seems this is not working as expected

> Given I am on the View Plan page
> AND Given there are Goals for Now in the Plan without Steps
> When I click Agree Plan
> Then the View Plan page is displayed with appropriate Errors

Is this expected to work as part of this ticket? Also no error when there is no goal and you can still agree to plan.

pipwilson:
No, the task I'm doing at the moment will validate that there are goals and that they have steps - you shouldn't be able to get to the Agree Plan page at all without those things being true.

pk2453 (approved):

PR #103, opened 2024-08-19
SP2-610 add e2e tests for agree plan

pk2453:


integration_tests/e2e/agree-plan.cy.ts
```diff
@@ -0,0 +1,70 @@
+describe('Agree plan', () => {
+  beforeEach(() => {
+    cy.createSentencePlan().then(planDetails => {
+      cy.wrap(planDetails).as('plan')
+      cy.openSentencePlan(planDetails.oasysAssessmentPk)
+      cy.get('.govuk-button').contains('Agree plan').click()
```
pipwilson, line 10: Once the gating of Agree is in place the tests will start failing because the Plan doesn't have any Goals which contain Steps. I suggest you use the DataGenerator to set that up here, and also create unhappy path tests for 1) plan with no goals and 2) plan with goals but no steps.

pk2453: I think this should be done when gating of Agree is in place. Its based on current state of main branch.

pipwilson: OK

pipwilson (approved):
LGTM :shipit:

PR #105, opened 2024-08-19
Changed IDs to fix autocomplete

Thomas-Geraghty:
- The static, an autocomplete field, both used the same ID
- It seems like, when the page load,  Cypress would begin running tests and targeting the field with id `goal-input-autocomplete`. 
- This was problematic because the field would be deleted from the DOM and replaced with the autocomplete version of the field, but Cypress would still attempt to type in the old, removed field - causing the test failure.


alexw-moj (approved):
LGTM

PR #109, opened 2024-08-22
Transform and validate middleware

Thomas-Geraghty:
- **Split up the validation middleware into two separate middleware**
  - transformRequest - Converts a plain object into it's DTO/Class instance
  ```
   transformRequest({
      body: EditGoalPostModel,
    }),
    ```
  - validateRequest - Simplified validateRequest to just perform validation on DTO/class instances, used like
  ```
  validateRequest(),
  ```
  - **Note**: Chain validateRequest after transformRequest, as validation rules are defined on our DTOs.

- **Moved validateRequest out of the routes.ts files, this always seemed like a bad place for it, so I'm happy now all the logic is held inside the controller**
- **Added 'runMiddlewareChain function to run through a chain of middleware in the same fashion Express does**

### Note - decent amount of changes in this PR, but they are all broken up commit-by-commit as usual, so it might be easier to review it that way!


server/middleware/validationMiddleware.ts
```diff
@@ -33,35 +26,23 @@ function simplifyValidationErrors(errorsList: ValidationError[]): Record<string,
...
-      req[DataSources.PARAMS] = plainToInstance(data.params, req[DataSources.PARAMS])
-      req.errors[DataSources.PARAMS] = getValidationErrors(req.params, data.params)
-    }
-
-    if (data.query) {
-      req[DataSources.QUERY] = plainToInstance(data.query, req[DataSources.QUERY])
-      req.errors[DataSources.QUERY] = getValidationErrors(req.query, data.query)
-    }
+    Object.values(RequestDataSources).forEach(source => {
+      if (req[source] && req[source].constructor && req[source].constructor.name !== 'Object') {
```
pk2453: can write this as ?
`req[source]?.constructor?.name !== 'Object'`

Thomas-Geraghty: Nice one @pk2453 , means I can rewrite this whole if statement to be a bit cleaner
```
req.errors[source] = req[source]?.constructor?.name !== 'Object' 
  ? getValidationErrors(req[source]) 
  : {}
```

Thomas-Geraghty: Resolved in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/109/commits/ef0af3f35c57351e4c6a00c0eace1aff1eb38e3f

pk2453 (approved):

PR #113, opened 2024-08-24
Interpolate and cleanup

Thomas-Geraghty:
- Switch all currently used views to use (excluding steps, in it's own PR as part of add-steps) to use the new interpolate function 
- Moved all hardcoded text strings into locale files, put variables in place
- Bit of general cleanup, removed some unnecessary 'isPageHeading' attributes, unused controller/views, small fix for agree-plan page.

Note-  as usual, all broken up by commit, might be easier to review that way.


server/routes/aboutPop/locale.json
```diff
@@ -1,6 +1,11 @@
 {
   "en": {
-    "title": "About {POP_NAME}",
+    "page": {
+      "title": "About {{ subject.givenName }} - {{ applicationName }}"
```
pipwilson: This should just say `About - {{ applicationName }}`

Thomas-Geraghty: Nice one, couldn't see anything on the spreadsheet for it, but this makes sense as we're doing no PII in the titles

Thomas-Geraghty: Resolved https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/113/commits/b4b6e2ceaf7c4fe8234609879588600eb174c1b4

pipwilson:
I think this all looks good but I think the "create step" page is missing.

pipwilson (approved):
LGTM. Big improvement.

PR #115, opened 2024-08-26
Setup common locale and preload pop-data into views

Thomas-Geraghty:
- Added a `mergeDeep` function that deeply merges objects/arrays. I thought about using a library for this, but it feels a bit unnecessary to add another library just for a single function
- Added a `commonLocale.json` which holds common locale strings for headers, footers, other 'common' strings used in components
    - Common locale can be accessed in views through `locale.common.`
-  Added a res.render proxy that preloads the common-locale, and popData into the view data.
    - This could likely be extended to the referenceData too, and the practitioner details data too.
- Removed all the unnecessary popData calls in controllers now that the data is loaded into the view as mentioned above

## Note - as usual, broken up by commit. Based on top of the #113 PR.


Thomas-Geraghty:
The commit history for this PR starts at https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/115/commits/d236133eead7eeadc56e18694cb154bd8312ac70

pipwilson (approved):
LGTM

PR #117, opened 2024-08-27
Update step types and StepController to use single actor

Thomas-Geraghty:
- Updated types to use single actor
- Added StepStatus enum
- Updated StepController to save single actor (getting replaced as part of add-steps PR)


pipwilson (approved):
Yes

PR #120, opened 2024-08-28
Fix issues with locale

Thomas-Geraghty:
- Remove a bunch of unneeded interpolate calls, this seems to have been leftover from a merge
- Moved set's for create-goal into the content block to fix locale not being interpolated correctly
- Update title setting to work correctly


pipwilson (approved):
Yes.

PR #122, opened 2024-08-29
Disable checking of date in cypress test

Thomas-Geraghty:
- Removed in new designs
- Broken today due to Aug having 31 days, Feb having 28


pipwilson (approved):
Great 🚀

PR #124, opened 2024-08-29
SP2-575: Adds cypress tests for Adding Steps

pipwilson:


integration_tests/e2e/add-steps.cy.ts
```diff
@@ -0,0 +1,174 @@
...
+describe('Add Steps', () => {
+  const addStep = new AddSteps()
+
+  const selectStepDescriptionByIndex = (index: number) => {
+    return cy.get(`table.goal-summary-card__steps .govuk-table__body > :nth-child(${index}) > :nth-child(1)`)
+  }
+
+  const selectStepActorByIndex = (index: number) => {
+    return cy.get(`table.goal-summary-card__steps .govuk-table__body > :nth-child(${index}) > :nth-child(2)`)
+  }
```
Thomas-Geraghty: Is it possible to move these out of the `describe('Add steps,()...)` function? Still in this file, just not defining the function within the test itself.

integration_tests/e2e/add-steps.cy.ts
```diff
@@ -0,0 +1,174 @@
...
+        })
+      })
+    })
+
+    it('Add one step', () => {
+      cy.url().should('include', '/add-steps')
+      const stepDescription = 'This is the first step'
+      const stepActor = 'Sam'
+      addStep.addStepAutocompleteText(1, stepDescription)
+      addStep.selectStepActor(1, stepActor)
```
Thomas-Geraghty: Any reason you didn't use the DataGenerator.generateStep() for generating some of this info? could have done
```
const step = DataGenerator.generateStep()

addStep.addStepAutocompleteText(1, step.description)
addStep.selectStepActor(1, step.actor)
```

No real difference in the end result, but it just means that, like in later steps, if you're creating multiple steps you don't need to make multiple step related variables

pipwilson: because I forgot it existed. I will do that, much better.

pipwilson: Addressed in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/124/commits/3e866574dd4374e959a1bf04f0401f6c8616474a

integration_tests/e2e/add-steps.cy.ts
pipwilson: Addressed in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/124/commits/3e866574dd4374e959a1bf04f0401f6c8616474a

Thomas-Geraghty (approved):

PR #125, opened 2024-08-30
SP2-274-Bugs - Fix styling issues

Thomas-Geraghty:
- Make autocomplete appear if 2 characters entered rather than 3
- Make autocomplete hide if no results found
- Make autocomplete show a selection hint at the top of the selection list
image

- Make add-steps fields use more of it's container

### DIFF
--- a/server/views/pages/add-steps.njk
+++ b/server/views/pages/add-steps.njk
@@ -13,6 +13,15 @@
     goal: data.goal
 }) %}
 
+{% block head %}
+    {{ super() }}
+    <style nonce="{{ cspNonce }}">
+        :root {
+            --autocomplete-menu-content: "{{ locale.step.description.autocompleteHint }}";
+        }
+    </style>
+{% endblock %}
+
 {% block pageTitle %}
     {% if not errors %}
         {{ locale.page.title }}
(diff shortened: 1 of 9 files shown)


server/views/pages/add-steps.njk
```diff
@@ -13,6 +13,15 @@
     goal: data.goal
 }) %}
 
+{% block head %}
+    {{ super() }}
+    <style nonce="{{ cspNonce }}">
```
alexw-moj, line 18: Why is this here, and not coming from a `css` file?

Thomas-Geraghty: Good question! Our CSS is statically built as part of our build process, converting the SCSS into a minified CSS file. Problem is, here we need to add strings into the CSS but have them be dependent on what page the user is on. 

We've added this :before element to the autocomplete menu, which shows a nice little hint as to what the user needs to do within this element. 
```
.autocomplete__menu {
  &:before {
    content: var(--autocomplete-menu-content);
    display: block;
    background-color: govuk-colour("light-grey");
    padding: govuk-spacing(1);
    border-bottom: 1px solid govuk-colour("mid-grey");
    font-weight: bold;
  }
```

We could hardcode the `content: var(--autocomplete-menu-content);` property, sure, but then we've got 2 problems:

1. We've got multiple autocomplete menu's across the app, some reference steps, some reference goals, so we'd need to support multiple strings dependent on what the user is interacting with
2. We've also got the lingering issue of multi-lang support, which, if we hardcoded the string, wouldn't work correctly when the user changes language.

A bit of a cheeky hack to solve these problems is instead to create a CSS variable, `--autocomplete-menu-content` and have that dynamically configured through our nunjucks template, and then use that variable within our CSS file. That allows us to then dynamically change the text based on page content and lang.

alexw-moj (approved):
LGTM

PR #126, opened 2024-08-30
SP2-611 add dynamic text to plan-summary page

pk2453:


server/routes/plan-summary/locale.json
```diff
@@ -13,6 +13,9 @@
       "changedGoal": "You changed a goal in {{ subject.possessiveName }} plan.",
       "removedGoal": "You removed a goal from {{ subject.possessiveName }} plan."
     },
+    "agreePlan": "{{ subject.givenName }} agreed to their plan on {{ plan.agreementDate }}",
+    "notAgreePlan": "Plan created on {{ plan.creationDate }}",
+    "lastUpdatedPlan": "Last updated on {{ plan.updatedDate }} by ---",
```
Thomas-Geraghty, line 18: Does the --- here need replacing with a variable? is it the practitioner's name that needs putting here?

pk2453: Yes I think this will require backend update. So just left as placeholder but ATM this message will never be displayed.

Thomas-Geraghty: Ah yeah, great point - we can't use the practitioner details from the handover, it'll have to be something passed when we update the plan/goals and then stored. I'll approve for now and we can address this later :)

Thomas-Geraghty (approved):

PR #136, opened 2024-09-10
SP2-698 Add practitioner info to API calls 

pipwilson:
This PR is not reviewable by going through each commit.

This change moves the RequestServices to be attached to each Express.Request object so that they can use an individual's authorisation token, rather than being a centralised pool of services which use the global application authorisation.

 The majority of this change is in `server/data/hmppsAuthClient.ts` - many of the other changes are the consequence of the move in services and the change to the constructors which means that routes and tests have all needed changing as well.


server/data/hmppsAuthClient.ts
```diff
@@ -19,10 +17,10 @@ function getSystemClientTokenFromHmppsAuth(username?: string): Promise<superagen
 
   const grantRequest = new URLSearchParams({
     grant_type: 'client_credentials',
-    ...(username && { username }),
+    username, // todo
```
alexw-moj, line 21: Is there something still `todo` here?

server/services/sessionService.ts
```diff
@@ -11,15 +12,19 @@ export default class SessionService {
...
       throw Error(e)
     }
   }
 
-  private getToken = () => this.request.user?.token
+  getToken = () => this.request.session.token
+
+  setToken = (token: Token) => {
+    this.request.session.token = token
+  }
```
Thomas-Geraghty: This could probably just be moved into hmppsAuthClient, as it's super unlikely we'll need to access this token elsewhere.

server/data/hmppsAuthClient.ts
```diff
@@ -19,10 +17,10 @@ function getSystemClientTokenFromHmppsAuth(username?: string): Promise<superagen
...
   }).toString()
 
-  logger.info(`${grantRequest} HMPPS Auth request for client id '${config.apis.hmppsAuth.systemClientId}''`)
+  logger.info(`${grantRequest} HMPPS Auth request for client id. '${config.apis.hmppsAuth.systemClientId}''`)
```
alexw-moj: Why the double single quotes at the end?

server/data/hmppsAuthClient.ts
pipwilson: Not in this PR but in the overall story yes - need to make sure we're passing through all of the correct information to the backend and wanted to leave myself a marker so I'd remember where it needed to be changed.

server/data/hmppsAuthClient.ts
pipwilson: It was like this on main - I can tidy that.

pipwilson: Fixed in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/136/commits/5fb51a21fa85dc9f0b8cc486751e895965630443

server/services/sessionService.ts
pipwilson: Fixed in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/136/commits/5fb51a21fa85dc9f0b8cc486751e895965630443

Thomas-Geraghty (approved):

PR #139, opened 2024-09-12
Adds authenticated user's ID to the username to be stored by the API

pipwilson:
We need to store both the ID and the name of the authenticated user at the backend, so this PR sends the user GUID through concatenated with the user's name in the username field. We can't add another field because of various external authentication validation issues.

### DIFF
--- a/server/data/hmppsAuthClient.test.ts
+++ b/server/data/hmppsAuthClient.test.ts
@@ -7,7 +7,7 @@ import mockReq from '../testutils/preMadeMocks/mockReq'
 
 jest.mock('./tokenStore/redisTokenStore')
 
-const username = 'Dr.+Benjamin+Runolfsdottir'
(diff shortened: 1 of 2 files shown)


server/data/hmppsAuthClient.test.ts
```diff
@@ -7,7 +7,7 @@ import mockReq from '../testutils/preMadeMocks/mockReq'
 
 jest.mock('./tokenStore/redisTokenStore')
 
-const username = 'Dr.+Benjamin+Runolfsdottir'
+const username = 'a23ccacf-7160-4431-9b4d-c560be9c9f5c%7CDr.+Benjamin+Runolfsdottir'
```
Thomas-Geraghty, line 10: Does this actually match how we expect the data to appear?  its `${identifier}|${username}` right?

pipwilson: Yes, this is the ID and Display Name from the Principal separated by a pipe (%7C once URL-encoded for use here)

Thomas-Geraghty (approved):

PR #147, opened 2024-09-18
SP2-657 Add ability to make a Goal as Achieved

pipwilson:
This PR adds two new controllers and two new pages. 

1) confirm that goal has been achieved
2) display achieved goal

Although it allows a Goal to marked as Achieved it does not include the ability to make a note at the same time - this will need to be added in a subsequent PR once SP2-594 is complete.


server/@types/NewGoalType.ts
```diff
@@ -1,6 +1,9 @@
+import { GoalStatus } from './GoalType'
+
 export type NewGoal = {
-  title: string
-  areaOfNeed: string
+  title?: string
```
Thomas-Geraghty: Doesn't a NewGoal always need a title?

pipwilson: the matching API change ensures that there's a submitted title when creating a new one, but basically I wanted to be able to submit a partial NewGoal in order to do updates against a known UUID without having to retrieve and submit all the data again.

Thomas-Geraghty: It might instead be better to use the `Partial<Goal>` type then, or `Partial<NewGoal>`, whichever your fancy

pipwilson: done in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/147/commits/f17ff510a0cf97caeaabf4526cb4a7f6eea6de47

Thomas-Geraghty (approved):

PR #155, opened 2024-10-02
SP2-819 Update plan summary page text

izzyL-DJ:
Updated plan summary page with better information:
Text to display when no current goals
Text to display when no future goals

### DIFF
--- a/server/views/pages/plan-summary.njk
+++ b/server/views/pages/plan-summary.njk
@@ -160,21 +160,22 @@
                 } if removedGoals.length > 0]
             }) }}
 
-            <ol id="goal-list" class="goal-list govuk-list govuk-list--number">
-                {% if data.type == 'future' %}
-                    {% set goals = futureGoals %}
-                    {% set goalType = 'future' %}
-                {% elseif data.type == 'achieved' %}
-                    {% set goals = achievedGoals %}
-                    {% set goalType = 'achieved' %}
-                {% elseif data.type == 'removed' %}
-                    {% set goals = removedGoals %}
-                    {% set goalType = 'removed' %}
-                {% else %}
-                    {% set goals = currentGoals %}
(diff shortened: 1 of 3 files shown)


server/views/pages/plan-summary.njk
```diff
@@ -132,7 +132,6 @@
             {% elseif data.plan.agreementStatus === 'DO_NOT_AGREE' %}
                 <p class="govuk-body">{{ locale.notAgreePlan }}</p>
             {% elseif data.plan.agreementStatus === 'DRAFT' %}
```
pipwilson: I think you can remove this elseif line now that there's no output with this agreementStatus

server/views/pages/plan-summary.njk
```diff
@@ -177,6 +176,18 @@
...
 
+                {% if goalType == 'current' and goals.length == 0 %}
+                    <p>{{ locale.ifNoCurrentGoals.text }}</p>
+                    <ul>
+                        <li><a href="/create-goal/accommodation">{{ locale.ifNoCurrentGoals.action1 }}</a></li>
+                        <li><a href="/about-pop">{{ locale.ifNoCurrentGoals.action2 }}</a></li>
+                    </ul>
+                {% endif %}
+
+                {% if goalType == 'future' and goals.length == 0 %}
```
pipwilson: Can you turn this into an elseif so that there's only one `if` statement in this new code?

pipwilson:
Now that I've looked at the markup again I think we should pull the `if..else` block on lines 164-176 to above the `<ol>` and then wrap that whole `<ol>....</ol>` in an {% if goals.length > 0 %} ... {% endif %} - that then means none of that code inside the `ol` needs to be processed and makes our following if statement make more sense too.

pipwilson (approved):

PR #161, opened 2024-10-11
SP2-818 Change URL

izzyL-DJ:
Change the url from plan-summary to plan as per design instructions


integration_tests/pages/plan.ts
```diff
@@ -1,4 +1,4 @@
-export default class PlanSummary {
+export default class Plan {
```
pipwilson: Let's call this class PlanOverview and the file plan-overview.ts because we know plan-summary is coming

integration_tests/e2e/achieve-goal.cy.ts
```diff
@@ -1,10 +1,10 @@
 import DataGenerator from '../support/DataGenerator'
 import { PlanType } from '../../server/@types/PlanType'
 import { Goal } from '../../server/@types/GoalType'
-import PlanSummary from '../pages/plan-summary'
+import Plan from '../pages/plan-overview'
```
pipwilson: for consistency you should refactor this to be named PlanOverview. Ditto in `plan-overview.cy.ts`.

pipwilson (approved):
LGTM

PR #165, opened 2024-10-16
SP2-869 Updated tests with more specificity in mind, after the recent url change

izzyL-DJ:


integration_tests/e2e/agree-plan.cy.ts
```diff
@@ -36,7 +36,7 @@ describe('Agree plan', () => {
       it('Do not allow access to agree-plan for plan with no goals', () => {
         cy.visit(`/agree-plan`)
 
-        cy.url().should('include', '/plan')
+        cy.url().should('eq', 'http://localhost:3000/plan')
```
pipwilson: so the tests don't always run under "localhost" e.g. on CircleCI but using `eq` and `/plan` should be fine for this and the one below

pipwilson (approved):

PR #177, opened 2024-10-22
SP2-836

izzyL-DJ:
Added context message for plan overview page to display when last the plan had been updated

### DIFF
--- a/server/utils/nunjucksSetup.ts
+++ b/server/utils/nunjucksSetup.ts
@@ -90,10 +90,6 @@ export default function nunjucksSetup(app: express.Express, applicationInfo: App
     return new Date(date).toISOString().substring(0, 10)
   })
 
-  njkEnv.addFilter('formatISODateWithSeconds', date => {
-    return new Date(date).toLocaleString()
-  })
-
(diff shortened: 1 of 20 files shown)


server/utils/nunjucksSetup.ts
```diff
@@ -90,6 +90,10 @@ export default function nunjucksSetup(app: express.Express, applicationInfo: App
     return new Date(date).toISOString().substring(0, 10)
   })
 
+  njkEnv.addFilter('formatISODateWithSeconds', date => {
```
pipwilson: I don't think we need this any more, can you take it out please?

server/views/pages/plan.njk
```diff
@@ -110,13 +111,15 @@
             {% endif %}
 
             {% if data.plan.agreementStatus === 'AGREED' %}
-                <p class="govuk-body">{{ locale.agreePlan }}</p>
+{#                {% if data.plan.updatedDate > data.plan.agreementDate %}to be finished at a later date#}
```
pipwilson: can you replace the "to finished..." text with `TODO: SP2-836` please? IntelliJ will auto-detect the TODO and it means we can search the codebase for incomplete stuff like this.

pipwilson (approved):

PR #180, opened 2024-10-23
SP2-884 Adds support for adding goal note to update goal

pipwilson:
This wraps the array of NewStep objects when creating or updating Steps in a NewGoal object so that we can optionally send a note with the steps without having to rewrite or duplicate chunks of the backend.

### DIFF
--- a/server/routes/update-goal/UpdateGoalController.ts
+++ b/server/routes/update-goal/UpdateGoalController.ts
@@ -43,7 +43,7 @@ export default class UpdateGoalController {
         if (goal.steps.some((step: StepModel, index) => step.uuid !== steps[index].uuid)) {
             throw createError(400, 'different steps were submitted')
         }
-    
+
         const updatedSteps: NewStep[] = goal.steps.map((value, index) => {
             return {
                 description: value.description,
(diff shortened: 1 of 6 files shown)


server/routes/update-goal/UpdateGoalController.ts
```diff
@@ -53,8 +51,28 @@ export default class UpdateGoalController {
...
+  }
+
+  private saveAndRedirect = async (req: Request, res: Response, next: NextFunction) => {
+    const { uuid } = req.params
+    const { steps } = req.body
+    const note = req.body.moreDetail
+
+    try {
+      await this.updateSteps(req, uuid, steps)
+      await this.addNoteToGoal(req, uuid, note)
```
alexw-moj: Was thinking about this, as one may succeed and one may fail, as the steps may be updated but without the note.

pipwilson: hm yes. also doing it like this means we create two new versions, which is not desirable. I think we will need to update the `updateStep` function in GoalController to take an optional note - do you agree?

pipwilson: Resolved in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/180/commits/627a4b1fd89717c29d050b0f4e426e2b71b4cb2b

alexw-moj (approved):

PR #188, opened 2024-11-01
SP2-648 Update Goals and Steps cypress tests

izzyL-DJ:


integration_tests/e2e/update-goal.cy.ts
```diff
@@ -64,6 +62,18 @@ describe('Update goal', () => {
...
     })
-    // it('') //should save and view notes
+
+    it('Can save and view notes attached to a goal', () => {
+      const planOverview = new PlanOverview()
+      planOverview.agreePlan()
+      cy.url().should('satisfy', url => url.endsWith('/plan')) // check we're back to plan-overview
+      cy.contains('a', 'Update').click() // click update link
+      cy.url().should('include', '/update-goal') // check url is update goal
+      cy.get('#more-detail').type('A goal update note') // type example note in the field
```
pipwilson: If you turn the value of the note here into a const and then also use that when you're checking the value on line 76 then that would be good 👍

pipwilson (approved):
LGTM

PR #190, opened 2024-11-01
Bugfix for update goal save

alexw-moj:


server/routes/update-goal/UpdateGoalController.test.ts
```diff
@@ -117,6 +117,16 @@ describe('UpdateGoalController', () => {
       expect(next).toHaveBeenCalledWith(new Error('different steps were submitted'))
     })
 
+    it('should redirect to the plan page when saved with no steps and no note', async () => {
```
pipwilson, line 120: I'm not sure this is right - should it not throw a validation error asking for something to be added? Can we check with UX?

server/routes/update-goal/UpdateGoalController.ts
```diff
@@ -57,10 +57,12 @@ export default class UpdateGoalController {
 
     const goalData: Partial<NewGoal> = {
       steps: updatedSteps,
-      note,
+      note: note || null,
```
pipwilson, line 60: this works but with the backend changes should not be necessary any more. OK to keep it if you prefer!

pipwilson (approved):

PR #196, opened 2024-11-06
SP2-915 View Plan Overview for Countersigning

pipwilson:
This PR redirects all users with an accessMode of READ_ONLY to the /plan URL when they try to access any URL (or back to sign-in if they haven't authorised yet). 

In PlanOverviewController the new countersign.njk template is rendered for a user with the READ_ONLY access mode. This removes all buttons except "Return to OASys" and removes all other goal/plan action links.

Reviewers can start with `server/middleware/authorisationMiddleware.ts` and `server/routes/planOverview/PlanOverviewController.ts`


alexw-moj (approved):
LGTM

PR #197, opened 2024-11-06
SP2-912 - Agree Plan page amendments 

izzyL-DJ:
Update to Agree Plan page:

- Change radio options text
- Add hint to radio option 3
- Change error messages for conditionals

### DIFF
--- a/integration_tests/pages/plan-overview.ts
+++ b/integration_tests/pages/plan-overview.ts
@@ -4,9 +4,9 @@ export default class PlanOverview {
   }
 
   agreePlan = () => {
-    cy.get('button').contains('Agree plan').click()
+    cy.get('button').contains('Save').click()
     cy.get('label').contains('Yes, ').click()
-    cy.get('button').contains('Agree plan with').click()
+    cy.get('button').contains('Save').click()
   }
 
   clickUpOnSummaryCard = (value: number) => {
--- a/integration_tests/pages/plan-overview.ts
+++ b/integration_tests/pages/plan-overview.ts
@@ -4,7 +4,7 @@ export default class PlanOverview {
   }
 
   agreePlan = () => {
-    cy.get('button').contains('Save').click()
+    cy.get('button').contains('Agree plan').click()
     cy.get('label').contains('Yes, ').click()
     cy.get('button').contains('Save').click()
   }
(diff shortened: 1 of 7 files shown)


server/routes/agree-plan/locale.json
```diff
@@ -9,24 +9,30 @@
       "hint": "{{ subject.givenName }} must answer this question, or you must record why {{ subject.givenName }} could not."
     },
     "agreePlanOptions": {
-      "yes": "Yes, {{ subject.givenName }} agrees",
-      "no": "No, {{ subject.givenName }} does not agree",
+      "yes": "Yes, I agree",
+      "no": "No, I do not agree",
+      "couldNotAnswerHint": "Share this plan with {{ subject.givenName }} next time you see them",
```
pipwilson: there should be a full stop at the end of this line

pipwilson (dismissed):
LGTM

integration_tests/pages/plan-overview.ts
```diff
@@ -4,9 +4,9 @@ export default class PlanOverview {
   }
 
   agreePlan = () => {
-    cy.get('button').contains('Agree plan').click()
+    cy.get('button').contains('Save').click()
```
pipwilson: Ah this change is wrong. The first button is the one on the Plan Overview page so should remain as "Agree plan"

server/utils/nunjucksSetup.ts
```diff
@@ -13,7 +13,7 @@ export default function nunjucksSetup(app: express.Express, applicationInfo: App
   app.set('view engine', 'njk')
 
   app.locals.asset_path = '/assets/'
-  app.locals.applicationName = 'Sentence plan'
+  app.locals.applicationName = 'Sentence Plan'
```
pipwilson: no, sorry this is a miscommunication. It is supposed to be "plan" - please revert this change.

pipwilson:
your e2e tests for update-goal fail because the test goes to /agree-plan manually. This is already handled by the PlanOverview.agreePlan call so you can remove `cy.visit('/agree-plan')` from line 21 of update-goal.cy.ts and then the test will pass - it was only luck it was passing previously because the buttons had the same text on them!

integration_tests/pages/plan-overview.ts
pipwilson: fixed in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/197/commits/378e56a8e70ec20ab998bba36f11e96f467ffe92

pipwilson (approved):
yay

PR #202, opened 2024-11-08
SP2-923 Create View Plan History Page

izzyL-DJ:


server/views/partials/plan-header/plan-header-mini.njk
```diff
@@ -17,6 +17,9 @@
                     <dt>{{ locale.common.planHeader.mini.dateOfBirthLabel }}</dt>
                     <dd>{{ data.popData.dateOfBirth }} </dd>
                 </div>
+                <div>
```
pipwilson: we don't need this yet, and it doesn't render properly on all pages, let's take this out.

server/views/pages/plan-history.njk
```diff
@@ -0,0 +1,35 @@
+{% from "govuk/components/details/macro.njk" import govukDetails %}
+{% extends "../partials/layout.njk" %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
```
pipwilson, line 2: Move this line up so all the govuk components are next to each other

server/views/pages/plan-history.njk
```diff
@@ -0,0 +1,35 @@
...
+{% endblock %}
+
+{% set mainClasses = "app-container govuk-body" %}
+
+{% block content %}
+<div class="govuk-grid-column-full">
+    <div class="govuk-grid-column-two-thirds">
+        <h1 class="govuk-heading-l govuk-!-margin-bottom-2">{{ locale.mainHeading.title }}</h1>
+    </div>
+    <div class="govuk-grid-column-one-third">
```
pipwilson: Drop these out for now - we can add them back in the relevant stories, otherwise we're bundling too much in here and the review will take forever

integration_tests/e2e/plan-history.cy.ts
```diff
@@ -0,0 +1,14 @@
...
+    cy.createSentencePlan().then(planDetails => {
+      cy.wrap(planDetails).as('plan')
+      cy.openSentencePlan(planDetails.oasysAssessmentPk)
+    })
+    cy.visit('/plan-history')
+  })
+
+  it('Display plan history page correctly on load', () => {
+    cy.url()
+    cy.get('.govuk-heading-l').contains('Plan history')
```
pipwilson: I think you should be able to use 'h1' as the selector in the `get` call.

pipwilson (approved):
LGTM

PR #206, opened 2024-11-11
Allows for viewing old readonly versions

pipwilson:
Reads the value of planVersion from sessionService and makes a request through a new planService method if planVersion exists.

Also adds a small bit of extra data integrity maintenance by checking the plan.readOnly property when rendering the Plan Overview and redirecting to the read-only view if so. This is so that users visiting old plan versions with READ_WRITE role can't make edits.

### DIFF
--- a/integration_tests/e2e/countersign.cy.ts
+++ b/integration_tests/e2e/countersign.cy.ts
@@ -16,8 +16,8 @@ describe('View Plan Overview for READ_ONLY user', () => {
   })
 
   it('Visiting create-goal should fail', () => {
-    cy.visit('/create-goal/accommodation')
-    cy.url().should('include', '/plan')
+    cy.visit('/create-goal/accommodation', { failOnStatusCode: false })
+    cy.url().should('not.include', '/plan')
   })
 
   it('Should have a `Return to OASys` button and it should return the user to the OASys return URL', () => {
--- a/integration_tests/e2e/countersign.cy.ts
+++ b/integration_tests/e2e/countersign.cy.ts
@@ -1,4 +1,6 @@
 import { AccessMode } from '../../server/@types/Handover'
+import DataGenerator from '../support/DataGenerator'
+import { PlanType } from '../../server/@types/PlanType'
 
 describe('View Plan Overview for READ_ONLY user', () => {
   beforeEach(() => {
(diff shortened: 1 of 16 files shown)


server/@types/PlanType.ts
```diff
@@ -28,4 +28,5 @@ export type PlanType = {
   updatedBy: string
   agreementDate: string
   goals: Goal[]
+  readOnly: boolean
```
Thomas-Geraghty: What's the need for this? Isn't the `read-only` attribute determined by the `READ_ONLY` role coming from OASys  (maintained in the handover context)

pipwilson: Our spec for PlanType does have a readOnly value and I thought this was a good point at which to introduce it. If we're not going to use it we should probably take it out of our object type in the API as well.

pipwilson: removed in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/206/commits/057bb2e02687440e511407401ee7829f85c2667b

integration_tests/e2e/countersign.cy.ts
```diff
@@ -26,3 +28,33 @@ describe('View Plan Overview for READ_ONLY user', () => {
...
+      cy.wrap(planDetails).as('plan')
+      pk = planDetails.oasysAssessmentPk
+      cy.addGoalToPlan(planDetails.plan.uuid, DataGenerator.generateGoal()).then(goal => {
+        cy.addStepToGoal(goal.uuid, DataGenerator.generateStep())
+      })
+      cy.lockPlan(planDetails.plan.uuid)
+    })
+  })
+
+  it('Should have one button', () => {
```
alexw-moj: Think there's more work to do here, tests nothing about buttons.

server/middleware/authorisationMiddleware.ts
```diff
@@ -1,21 +1,30 @@
...
-      }
       return next()
     }
 
     req.session.returnTo = req.originalUrl
     return res.redirect('/sign-in')
   }
 }
+
+export function hasAccessMode(requiredAccessMode: AccessMode): RequestHandler {
```
alexw-moj: Can we rename this to `validateAccessMode`?

integration_tests/e2e/countersign.cy.ts
pipwilson: finished this as much as I can right now in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/206/commits/5d7fe1e6e2b9300569a9695c5e767ccf1dfc8554 - ideally I think it would sign out and then sign back in again to check the list twice but our sign-out doesn't quite work right so trusting that `addGoalToPlan` is I think as much as I can do now!

server/middleware/authorisationMiddleware.ts
pipwilson: renamed in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/206/commits/db72907e836bee0b129166ca960242c987f07028

alexw-moj (approved):

PR #208, opened 2024-11-13
SP2-407: Person Summary

pmcphee77:
Collapsed view
localhost_3000_about-pop-collapsed
Full view
localhost_3000_about-pop-full


Dockerfile
```diff
@@ -27,7 +27,7 @@ ENV GIT_BRANCH=${GIT_BRANCH}
 RUN apt-get update && \
     apt-get upgrade -y && \
     apt-get autoremove -y && \
-    apt-get install -y make python g++ curl && \
+    apt-get install -y make python g++ curl default-jre && \
```
Thomas-Geraghty: How come we're adding a java-runtime-environment to a node-based container?

assets/js/print.js
```diff
@@ -0,0 +1,34 @@
...
+  detailsElements.forEach(details => {
+    if (!details.open) {
+      details.open = true
+    }
+  })
+
+  // Wait for a short moment to ensure all accordions are open, then trigger print
+  setTimeout(() => {
+    window.print()
+  }, 500) // Adjust the delay if necessary
```
Thomas-Geraghty: I'm surprised that accordions don't print by default in an open state, this seems like an oversight by the govuk-frontend if this isn't the default behaviour.

assets/js/print.js
```diff
@@ -0,0 +1,34 @@
...
+}
+
+// Event listener for the print button
+if (document.getElementById('print-button')) {
+  document.getElementById('print-button').addEventListener('click', openAccordionsAndPrint)
+} else {
+  console.log('no print button')
+}
```
Thomas-Geraghty: Don't need to log the lack of something, this will trigger on nearly every page

Dockerfile
pmcphee77: Its because i needed to use wiremock and the dev environment uses docker. Im looking to see if we can still use wiremock, but start it outside of the container. Also discussing this with Alex

assets/js/print.js
pmcphee77: im thinking of taking the print facility out of this work as this isnt actually specified in the ticket

assets/js/print.js
pmcphee77: As above

server/routes/aboutPop/locale.json
```diff
@@ -3,9 +3,17 @@
     "page": {
       "title": "About - {{ applicationName }}"
     },
+    "header": {
+      "returnToOasysButton": "Return to OASys",
```
Thomas-Geraghty: I wonder if this should be moved to the common-locale?

pipwilson: Since this branch started this got added to commonLocale.json as `header.returnToOasysButton`.

server/routes/aboutPop/locale.json
```diff
@@ -17,6 +25,64 @@
...
+      },
+      {
+        "area": "Education and employment",
+        "crimNeedsKey": "educationTrainingEmployability",
+        "crimNeedsSubKey": "ete",
+        "assessmentKey": "employment_education",
+        "goalRoute": "employment-and-education"
+      },
+      {
+        "area": "Finance",
```
alexw-moj: We've changes this to `Finances` as part of the design.

server/utils/utils.ts
```diff
@@ -54,6 +62,119 @@ export function formatDate(date: string): string {
...
+  const lowScoring = all.filter(area => Number(area.criminogenicNeedsScore) <= 3)
+  const highScoring = all.filter(area => Number(area.criminogenicNeedsScore) > 3)
+  return { lowScoring, highScoring, versionUpdatedAt: assessment.metaData?.versionUpdatedAt } as AssessmentAreas
+}
+
+export const toHtml = (content: string): string => {
+  return content.replace('\n', '&#13;&#10;')
+}
+
+export const motivationText = (optionResult: string, forename: string): string => {
```
alexw-moj: Is this something we could do via the locale?

server/routes/aboutPop/locale.json
pmcphee77: Changed

assets/js/print.js
pmcphee77: Removed

assets/js/print.js
pmcphee77: removed

server/routes/aboutPop/locale.json
pmcphee77: changed

server/utils/utils.ts
pmcphee77: Removed function above as thats unused. Looking at using locales

server/utils/utils.ts
```diff
@@ -54,6 +62,119 @@ export function formatDate(date: string): string {
...
+        riskOfReoffending,
+        strengthsOrProtectiveFactors,
+        criminogenicNeedsScore: score,
+        goalRoute: area.goalRoute,
+      } as AssessmentArea
+    })
+    .sort((a, b) => 0 - (a.criminogenicNeedsScore > b.criminogenicNeedsScore ? 1 : -1))
+
+  const lowScoring = all.filter(area => Number(area.criminogenicNeedsScore) <= 3)
+  const highScoring = all.filter(area => Number(area.criminogenicNeedsScore) > 3)
```
alexw-moj, line 117: If the `criminogenicNeedsScore` is undefined, then areas get filtered out and won't display. 

BA says that every section should always be displayed.

assets/scss/local.scss
Thomas-Geraghty, line 1: We likely want to use the `govuk-frontend` mixins for any hardcoded, like so 
`@include govuk-responsive-padding(6)` 
or at least 
`padding-top: govuk-spacing(6)`.

If these values change in future, it'll cause a website to no longer match styling, so it's probably best to just do this the correct way from the start

package.json
```diff
@@ -17,7 +17,8 @@
     "int-test": "cypress run --config video=false",
     "int-test-ui": "cypress open --e2e --browser chrome",
     "clean": "rm -rf dist build node_modules stylesheets",
-    "rebuild": "npm run clean && npm i && npm run build"
+    "rebuild": "npm run clean && npm i && npm run build",
+    "wiremock": "npx wiremock --port 9095 --root-dir wiremock --local-response-templating"
```
Thomas-Geraghty: Do we maybe want to run wiremock instead through Docker? I've got some of the stuff already written to do this, and it'd likely be more appropriate.

server/@types/Handover.ts
```diff
@@ -1,9 +1,12 @@
+import { CriminogenicNeedsData } from './Assessment'
+
 export type HandoverContextData = {
   handoverSessionId: string
   principal: HandoverPrincipal
   subject: HandoverSubject
   assessmentContext: HandoverAssessmentContext
   sentencePlanContext: HandoverSentencePlanContext
+  criminogenicNeedsData?: CriminogenicNeedsData
```
Thomas-Geraghty, line 9: Is the `criminogenicNeedsData` field not always suppled? I think there's times where the contents might be empty/undefined, but I think the field might always be there

server/data/coordinatorApiClient.ts
```diff
@@ -0,0 +1,14 @@
...
+import HmppsAuthClient from './hmppsAuthClient'
+import logger from '../../logger'
+
+export default class CoordinatorApiClient {
+  constructor(private readonly authClient: HmppsAuthClient) {}
+
+  async restClient(info?: string): Promise<RestClient> {
+    if (info) logger.info(info)
+    const token = await this.authClient.getSystemClientToken()
+    return new RestClient('Coordinator Plan Client', config.apis.coordinatorApi, token)
```
Thomas-Geraghty:
```suggestion
    return new RestClient('Coordinator API Client', config.apis.coordinatorApi, token)
```

server/data/restClient.ts
```diff
@@ -72,6 +79,14 @@ export default class RestClient {
...
       return raw ? result : result.body
     } catch (error) {
+      if (handle500 && error.response?.status === 500) {
+        const warnings: ErrorSummaryItem[] = []
+        warnings.push({ text: errorMessageFor500 })
+        error.response.errors = warnings
+        logger.info('Handling 500')
+        return error.response
+      }
+      if (handle404 && error.response?.status === 404) return null
```
Thomas-Geraghty, line 89: Mind explaining this change? I don't fully understand it, and I'm cautious that it's changing the restClient which is used across multiple different services

pmcphee77: it allows optionall handling for 404s and 500s coming back from APIs, we have it in MmS. So we can put messages back when an assessment has not been found or delius is down etc. Seems a shame to just bomb out if only one of the apis returns a 404 or 500.

server/@types/Handover.ts
pmcphee77: im not sure if its always going to be supplied, so just handling nulls in case

package.json
pmcphee77: yeah im doing this now

server/utils/utils.ts
pmcphee77: yeah, i asked this question this morning, but if no needs score, how will we know what section to put them in? Low or High (theres only 2). They said that all these sections should have a needs score. So we will just need to wait and see

server/views/components/assessment/_assessment-area.njk
```diff
@@ -0,0 +1,81 @@
...
+                                <h3 class="govuk-heading-s">There are strengths or protective factors related to
+                                    this area</h3>
+                                <p class="govuk-body">{{ assessment.strengthsOrProtectiveFactors }}</p>
+                            </div>
+                        {% else %}
+                            <h3 class="govuk-heading-s">There are no strengths or protective factors related to this area</h3>
+                        {% endif %}
+                        <div class="assessment-score">
+                            <h3 class="govuk-heading-s">{{ assessment.title }} need score</h3>
+                            <p class="govuk-body">{{ assessment.criminogenicNeedsScore }} out of 10. (Scores above 3 are high-scoring.)</p>
```
alexw-moj: Scores aren't always out of 10.

package.json
pmcphee77: Done

server/data/coordinatorApiClient.ts
pmcphee77: done

server/views/components/assessment/_assessment-area.njk
pmcphee77: Ok - changed the component to take in an upper bound and set upper bounds on areas

server/utils/utils.ts
pmcphee77: Now using locales

server/utils/utils.ts
pmcphee77: Added an other section

assets/scss/local.scss
pmcphee77: As discussed with Tom, we can look at this separately

Dockerfile
pmcphee77: reverted this change, now using docker

Thomas-Geraghty: Great stuff, resolving

package.json
Thomas-Geraghty: Good stuff! closing

assets/scss/application.scss
```diff
@@ -23,4 +23,6 @@ $govuk-suppressed-warnings: 'govuk-typography-scale-14';
 @import "./components/add-steps-list";
 @import "./components/accessible-autocomplete";
 @import "./components/select";
+@import "components/needs-scores";
 @import './local';
+@import "node_modules/@ministryofjustice/frontend/moj/all";
```
alexw-moj: Is this necessary, given line 10?

pmcphee77: changed

server/views/pages/about-pop.njk
```diff
@@ -1,7 +1,15 @@
...
+{%- from "moj/components/banner/macro.njk" import mojBanner -%}
+{% from "../components/assessment/assessment.njk" import renderAssessment %}
+{% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
+
+{% set print = true %}
+{% set locale = interpolate(locale, {
+    subject: data.popData
+}) %}
 
 {% block pageTitle %}
```
alexw-moj, line 14: `locale.page.errorTitle` is used in the template, but not defined in the locale file.

alexw-moj (approved):

PR #215, opened 2024-11-14
SP2-918 bugfix update goal

alexw-moj:
Currently draft, as relates to https://github.com/ministryofjustice/hmpps-sentence-plan/pull/145

### DIFF
--- a/integration_tests/e2e/change-goal.cy.ts
+++ b/integration_tests/e2e/change-goal.cy.ts
@@ -110,6 +110,15 @@ describe('Change a goal', () => {
       cy.url().should('include', 'plan?status=updated&type=current')
     })
 
+    it('Should change goal with standard date and related areas of need are retained', () => {
+      // Modify data
+      cy.get('#date-selection-radio-2').check()
+
+      cy.contains('button', 'save').click()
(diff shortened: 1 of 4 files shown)


integration_tests/e2e/change-goal.cy.ts
```diff
@@ -110,22 +110,33 @@ describe('Change a goal', () => {
       cy.url().should('include', 'plan?status=updated&type=current')
     })
 
+    it('Should change goal with standard date and related areas of need are retained', () => {
+      // Modify data
+      cy.get('#date-selection-radio-2').check()
```
pipwilson: Here and in other radio/checkbox getters could you use the text value of the item please? It's really hard to tell what these opaque IDs are selecting (I know it's a legacy problem but we should try not to add to it).

alexw-moj: Added selector changes in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/215/commits/269aa021989ceb1f6f35d64682b1ef4da05cafa4

pipwilson (approved):
LGTM

PR #217, opened 2024-11-15
SP2-926 [Design review update] Updated plan history page 

izzyL-DJ:
to show different introductions when there iss no plan present and when there is, added cypress test for it

### DIFF
--- a/server/views/pages/plan-history.njk
+++ b/server/views/pages/plan-history.njk
@@ -29,7 +29,11 @@
 {% endblock %}
 
 {% block content %}
-    <p class='govuk-!-margin-top-5'>{{ locale.page.introduction }}</p>
+    {% if data.notes.length == 0 %}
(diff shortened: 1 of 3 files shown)


server/views/pages/plan-history.njk
```diff
@@ -29,15 +29,18 @@
 {% endblock %}
 
 {% block content %}
-    <p class='govuk-!-margin-top-5'>{{ locale.page.introduction }}</p>
+    {% if data.notes.length == 0 %}
+        <p class='plan-history-intro govuk-!-margin-top-5'>{{ locale.page.noPlanIntroduction }}</p>
```
pipwilson: if you move the if/endif to inside the <p> then you only need that line once rather than twice. Can you do that please? So code will look like
p
if
else
endif
p

pipwilson (approved):
LGTM

PR #223, opened 2024-11-19
SP2-407: Correct assessment api response structure to match api

pmcphee77:


server/utils/utils.ts
```diff
@@ -69,7 +69,7 @@ export const formatAssessmentData = (
   assessment: AssessmentResponse,
   areas: AssessmentAreaConfig[],
 ): AssessmentAreas => {
-  if (assessment === null) {
+  if (assessment === null || assessment === undefined || assessment.assessment === undefined) {
```
Thomas-Geraghty:
```suggestion
  if (!assessment.assessment) {
```
this could be written like this, right?

Thomas-Geraghty (approved):

PR #238, opened 2024-11-21
SP2-1035 Add lifestyle score to TBA

alexw-moj:
Temporary addition of the Lifestyle & Associates score to the Thinking Behaviours and Attitudes section.


server/utils/utils.test.ts
```diff
@@ -241,9 +241,11 @@ describe('format assessment data', () => {
             linkedtoReoffending: true,
             linkedtoRoSH: false,
             motivationToMakeChanges: undefined,
+            overallScore: '6',
             riskOfReoffending: undefined,
             riskOfSeriousHarm: undefined,
             strengthsOrProtectiveFactors: undefined,
+            subData: undefined,
```
Thomas-Geraghty: Any reason to include if it's undefined? Is the type too strong and enforcing it?

server/utils/utils.ts
```diff
@@ -88,6 +91,19 @@ export const formatAssessmentData = (
...
+        subData = {
+          upperBound: '6',
+          thresholdValue: '1',
+          criminogenicNeedsScore: crimNeeds.lifestyleAndAssociates.lifestyleOtherWeightedScore,
+        }
+        overallScore = Math.max(
+          Number(crimNeeds.thinkingBehaviourAndAttitudes.thinkOtherWeightedScore),
+          Number(subData.criminogenicNeedsScore),
+        )
+      }
```
Thomas-Geraghty, line 109: Null checks needed here? Seems like you might be able to cause an exception if `thinkingBehaviourAndAttitudes` was null for example, and then you tried turn null into a number (though maybe it can handle this)

alexw-moj: Added check for `thinkingBehaviourAndAttitude`. 

`Math.max(null, null)` = `0`

server/views/components/assessment/_assessment-area.njk
```diff
@@ -82,6 +80,25 @@
                                 <p>{{ locale.needScore.needScoreNotAvailable }}</p>
                             {% endif %}
                         {% endif %}
+
+                        {% if assessment.subData %}
+                            <br>
+                            <div class="assessment-score">
+                                <h3 class="govuk-heading-s">{{ locale.detail.lifestyleAndAssociatedNeedScore }}</h3>
+                                <p class="govuk-body">{{ assessment.subData.criminogenicNeedsScore }} out of {{ assessment.subData.upperBound }}. (Scores above
```
Thomas-Geraghty, line 88: Hate to be that guy buuuut, it looks like you've gone and switched to using the locale file - which is really good, but accidentally missed the `out of` string that is used in a few places 😅

alexw-moj: I kinda left that one cos the other PR is doing scores.

Thomas-Geraghty: I will accept this answer and put it on the other PR instead :)

Thomas-Geraghty (approved):

PR #253, opened 2024-11-22
Display threshold for subdata

pipwilson:


server/utils/utils.ts
```diff
@@ -100,7 +100,7 @@ export const formatAssessmentData = (
       ) {
         subData = {
           upperBound: '6',
```
alexw-moj, line 102: This probably shouldn't be hardcoded anymore.

alexw-moj (approved):

PR #255, opened 2024-11-22
Orders the Areas relative to their distance from the threshold 

pipwilson:
The real change here is in server/utils/utils.ts


server/utils/utils.ts
```diff
@@ -139,7 +139,11 @@ export const formatAssessmentData = (
         subData,
       } as AssessmentArea
     })
-    .sort((a, b) => 0 - (a.criminogenicNeedsScore > b.criminogenicNeedsScore ? 1 : -1))
+    .sort((a, b) =>
+      Number(b.criminogenicNeedsScore) - b.thresholdValue - (Number(a.criminogenicNeedsScore) - a.thresholdValue) > 0
```
alexw-moj, line 143: Can you add tests to make sure it's sorting things in the right order and scoring area?

pipwilson: The changes made to the existing tests in this PR should be doing that I think

alexw-moj (approved):

PR #257, opened 2024-11-25
Renames routes so they are all consistent

pipwilson:
The renames the routes to all use hyphens instead of a mix of hyphenated-names and camelCaseNames.

It also renames the AboutPopController to AboutPersonController.


server/routes/about-person/AboutPersonController.ts
```diff
@@ -2,7 +2,7 @@ import { NextFunction, Request, Response } from 'express'
 import locale from './locale.json'
 import { formatAssessmentData } from '../../utils/utils'
 
-export default class AboutPopController {
+export default class AboutPersonController {
```
alexw-moj, line 5: Url is just `/about` now, so just `AboutController` and `about` folder?

server/routes/about-person/routes.ts
```diff
@@ -1,9 +1,9 @@
 import { Router } from 'express'
-import AboutPopController from './AboutPopController'
+import AboutPersonController from './AboutPersonController'
 import URLs from '../URLs'
 
 export default function setupAboutPopRoutes(router: Router) {
-  const controller = new AboutPopController()
+  const controller = new AboutPersonController()
 
   router.get(URLs.ABOUT_POP, controller.get)
```
alexw-moj: `_POP`. Other function and variable names also use pop terminology too.

server/routes/about-person/AboutPersonController.ts
pipwilson: I thought about that, but I thought this was clearer, similar to how we've left Plan as PlanOverview so there's a bit more of a clue to the humans on what the controller does.

server/routes/about-person/routes.ts
pipwilson: My main goal was getting the file structure consistent. I'll rename ABOUT_POP though and have a look through the other `pop`s.

pipwilson:
Converted to draft until some of the other changes have come in - this change is not a priority right now.

PR #260, opened 2024-11-26
SP21-1040 Fixes the maths for reordering needs

pipwilson:
Reordering is based on the distance between the score and the threshold. This is performed inside the new `filterAndSortAreas` function in utils.ts.


server/utils/utils.ts
```diff
@@ -74,83 +74,80 @@ export const formatAssessmentData = (
...
-          score = area.upperBound
-        }
+  const all = Object.values(areas).map(area => {
+    let score
+    let linkedtoRoSH
+    let linkedtoReoffending
+    let subData: SubAreaData
+    let overallScore
+
+    if (Object.prototype.hasOwnProperty.call(crimNeeds, area.crimNeedsKey)) {
```
Thomas-Geraghty: Any reason for doing this using `Object.prototype.hasOwnProperty.call`? seems a bit odd school-y. 

Could do 
- `if (area.crimNeedsKey in crimNeeds)`
- `if (area.crimNeedsKey[area.crimNeedsKey])` (if the value can come out as false, add an undefined check)
- `if (Object.hasOwn(crimNeeds, area.crimNeedsKey)`

I can't imagine there'll be a prototype property collision with the the expected data on this object, so you're probably okay to do a less robust check like the first 2 examples

server/utils/utils.ts
```diff
@@ -74,83 +74,80 @@ export const formatAssessmentData = (
...
+  const highScoring = filterAndSortAreas(all, (score, threshold) => score > threshold)
   const other = all.filter(area => area.criminogenicNeedsScore === undefined)
   return { lowScoring, highScoring, other, versionUpdatedAt: assessment.lastUpdatedTimestampSAN } as AssessmentAreas
 }
 
+const filterAndSortAreas = (areas: AssessmentArea[], comparator: (score: number, threshold: number) => boolean) => {
+  return areas
+    .filter(area => comparator(Number(area.overallScore), area.thresholdValue))
+    .sort((a, b) => (Number(a.overallScore) - a.thresholdValue > Number(b.overallScore) - b.thresholdValue ? -1 : 1))
+}
```
Thomas-Geraghty: Because this always returns -1 or 1, will this cause odd shifting in the sort order when the calculated values for `a` and `b` are the same?

server/utils/utils.ts
pipwilson: I didn't write this line, so I don't know. This was a whitespace change. I can change it though.

server/utils/utils.ts
pipwilson: Yes. I'll add an extra check for score equality and sort by title.

server/utils/utils.ts
pipwilson: Fixed in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/260/commits/6be11af672980c9e4990cddececcf804bfd7efd8

server/utils/utils.ts
pipwilson: Fixed in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/260/commits/6be11af672980c9e4990cddececcf804bfd7efd8

Thomas-Geraghty (approved):

PR #276, opened 2024-11-29
SP2-407 Fixes sentence information durations

pipwilson:
Duration description are supposed to match the comment linked on this Jira story. This change supports displaying all combinations of years, months and days rather than just years and days.

### DIFF
--- a/server/utils/assessmentUtils.ts
+++ b/server/utils/assessmentUtils.ts
@@ -132,10 +132,42 @@ export const dateWithYear = (datetimeString: string): string | null => {
   return DateTime.fromISO(datetimeString).toFormat('d MMMM yyyy')
 }
 
-export const yearsAndDaysElapsed = (datetimeStringFrom: string, datetimeStringTo: string): string => {
+export const yearsAndDaysElapsed = (datetimeStringFrom: string, datetimeStringTo: string): any => {
   if (!datetimeStringFrom || isBlank(datetimeStringFrom)) return undefined
   if (!datetimeStringTo || isBlank(datetimeStringTo)) return undefined
-  const yearsDays = DateTime.fromISO(datetimeStringTo).diff(DateTime.fromISO(datetimeStringFrom), ['years', 'days'])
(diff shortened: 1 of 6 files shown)


server/utils/assessmentUtils.ts
```diff
@@ -132,10 +132,42 @@ export const dateWithYear = (datetimeString: string): string | null => {
...
+  } else if (yearsMonthsDays.years > 0 && yearsMonthsDays.days > 0) {
+    sentenceLengthstring = `${pluralise(yearsMonthsDays.years, locale.year)} ${locale.conjunction} ${pluralise(yearsMonthsDays.days, locale.day)}`
+  } else if (yearsMonthsDays.years > 0) {
+    sentenceLengthstring = `${pluralise(yearsMonthsDays.years, locale.year)}`
+  } else if (yearsMonthsDays.months > 0) {
+    sentenceLengthstring = `${pluralise(yearsMonthsDays.months, locale.month)}`
+  } else if (yearsMonthsDays.days > 0) {
+    sentenceLengthstring = `${pluralise(yearsMonthsDays.days, locale.day)}`
+  }
+
```
Thomas-Geraghty, line 172: this looks widely complex, mind if we just pop a comment on here describing how/what all of this does?

pipwilson: I can yes - but fundamentally it's just all the different combinations of some or all of years, months and days, using different characters to join them - thinking about it now there's probably some minor improvements (e.g. automatically providing the comma or "and" based on number of things to output), but you'd still need some form of case detection so it's probably an unnecessary complication.

There might be a nicer approach to handling this, but at least this one is straightforward? 🤷

pipwilson: Comment added in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/276/commits/ae0225723e75aad87b2366c351e8dc32c25d3041

Thomas-Geraghty (approved):

PR #278, opened 2024-12-02
Updated error message when plan is agreed with no steps on the plan page as per design instructions

izzyL-DJ:


pipwilson (approved):
LGTM

PR #279, opened 2024-12-02
SP2-700: Fixed steps being removed when pressing enter issue

JakeGoodwin22:
This PR fixes a bug that would result in a step being removed when the user pressed enter. The issue was that the form selects the first form action on keypress. I have tried to change this to an <a> tag with no luck, it just stopped working all together.

The solution which seems to work is to have a copy of the Add another step button at the top of the form, visually hidden from the user. I have ran the fix against Axe which doesn't have any accessibility issues.

### DIFF
--- a/integration_tests/e2e/add-steps.cy.ts
+++ b/integration_tests/e2e/add-steps.cy.ts
@@ -81,7 +81,7 @@ describe('Add Steps', () => {
       addStep.addStepAutocompleteText(2, secondStep.description)
       addStep.selectStepActor(2, secondStep.actor)
 
-      addStep.saveAndContinue()
+      addStep.saveAndContinuePressingEnter(2)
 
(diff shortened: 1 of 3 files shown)


integration_tests/e2e/add-steps.cy.ts
```diff
@@ -81,13 +81,21 @@ describe('Add Steps', () => {
       addStep.addStepAutocompleteText(2, secondStep.description)
       addStep.selectStepActor(2, secondStep.actor)
 
+      addStep.AddAnotherStepPressingEnter(2)
```
JakeGoodwin22, line 84: Extended a current test to show that pressing enter now has the same functionality as selecting the "add another step" button

Thomas-Geraghty (approved):

PR #280, opened 2024-12-02
SP2-1023

alexw-moj:
Adding Axe accessibility testing to the Cypress tests, and resolving various issues that were flagged.


server/views/pages/plan.njk
```diff
@@ -158,8 +158,11 @@
                 {
                     text: locale.subNavigation.removedGoals,
                     href: '/plan?type=removed',
-                    active: data.type === 'removed'
-                } if removedGoals.length > 0]
+                    active: data.type === 'removed',
+                    attributes: {
+                        'data-module': 'float'
+                    }
```
Thomas-Geraghty: what does this do? 🫨

alexw-moj: We were creating additional, empty, links in the sub-nav, so the last child was always the one that needed to float. As that's no longer true (thanks to `reject('string')` below), I have added something to the only one that needs to float.

See `assets/scss/local.scss`.

Thomas-Geraghty: this isn't really the intended usage for data attributes, it would instead be better to have the attribute based on either the content of the button, the link href, or just do a child nth selector to select the 4th child element and float it to the right.

alexw-moj: Amended to use the href in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/280/commits/bcc37e9acaaa75bb61c46a1316cb2996f51f009f

Thomas-Geraghty (approved):

PR #281, opened 2024-12-02
SP2-1053: Add clear option to add steps

JakeGoodwin22:
This change introduces a "Clear" option for when we only have a single step added to a goal. This does two things for us:

1. Is clearer to the user that they must have at least 1 step. They can clear the contents but must have one or more
2. Fixes a bug where removing all of the steps in a goal (when there were multiple) which would cause strange behaviour

Screenshot 2024-12-02 at 15 03 27

Screenshot 2024-12-02 at 15 03 47


Thomas-Geraghty:
I've not reviewed this yet but lets make sure we don't merge this in until PO have signed off on this change 😃

alexw-moj (approved):
LGTM

PR #284, opened 2024-12-03
SP2-1023 region

alexw-moj:
Moves the plan-header into the `main` landmark to resolve the [region](https://dequeuniversity.com/rules/axe/4.1/region) issue from Axe.


pipwilson:
It would have been super-useful to have had a bit of rationale about these changes in the PR description, but I eventually worked it out by trial and error.

Approving, but I'm unclear about why you've removed `@include govuk-width-container;` from `_plan-header.scss` - was it extraneous?

pipwilson (approved):
LGTM

PR #285, opened 2024-12-03
SP2-716: Stop defaulting the Actor field on 'Add steps' page

JakeGoodwin22:
Screenshot 2024-12-03 at 10 18 56


pipwilson (approved):
LGTM

PR #289, opened 2024-12-04
Moves the brackets for sentence duration into the calculator from the template

pipwilson:
If we leave the brackets in the template then when a sentence is missing an end date (which is a permitted state) you end up with `()` displaying on the About page.


izzyL-DJ (approved):
LGTM

PR #290, opened 2024-12-05
SP2-1052-Remove Add to plan link on removed goal

izzyL-DJ:
As the functionality for this isn't implemented yet, this link serves no purpose. So it's currently commented out with a reference to the ticket that implements this feature


pipwilson (approved):
LGTM

PR #291, opened 2024-12-05
Passes the OASys Return URL to the error page.

pipwilson:
The locale file for the error page relied on data being passed it in order to render the OASys Return URL, but nothing was being passed through. This PR fixes that.


izzyL-DJ (approved):
LGTM

PR #292, opened 2024-12-05
Adds app-container class to main element

pipwilson:
Before change:

Screenshot 2024-12-05 at 15 52 14

After change:

Screenshot 2024-12-05 at 15 53 05


izzyL-DJ (approved):
LGTM

PR #300, opened 2024-12-06
Switch to using `hmpps-monitoring` library

Thomas-Geraghty:
- Remove existing health check functions
- Switch to using new `hmpps-monitoring` library for performing more robust health checks


alexw-moj:
Is this going to cause our pods to cycle (probably for no reason) if one or more of its dependencies reports unhealthy?

server/config.ts
```diff
@@ -122,6 +118,7 @@ export default {
     },
     coordinatorApi: {
       url: get('COORDINATOR_API_URL', 'https://arns-coordinator-api-dev.hmpps.service.justice.gov.uk'),
+      healthPath: '/health/ping',
```
alexw-moj, line 121: Turned all of the above off in docker, but this one never reports DOWN...

Thomas-Geraghty: Did you turn off the wiremock docker container too?

Thomas-Geraghty, line 121: Apologies @alexw-moj my GitHub just messed up and deleted your comment. For future reference, it was a question along the lines of "I've disabled all the containers but this health check continued to pass".

My follow up comment was; did you ensure to stop the Wiremock container too? On our local development environment, it uses the wiremock for the coordinator.

Thomas-Geraghty:
> Is this going to cause our pods to cycle (probably for no reason) if one or more of its dependencies reports unhealthy?

Likely, but this is probably a wider issue and not related specifically to introducing this library

server/config.ts
alexw-moj: Lol, no I deleted it just as you posted, when I figured that out myself :)

alexw-moj:
> > Is this going to cause our pods to cycle (probably for no reason) if one or more of its dependencies reports unhealthy?
> 
> Likely, but this is probably a wider issue and not related specifically to introducing this library

Does the `isEnabled` option disable the healthcheck itself, or just whether it's used in the calculation of `UP` or `DOWN`? Looks like it just turns them off, but the calculation thing might be useful?

alexw-moj (approved):

Thomas-Geraghty:
>Does the isEnabled option disable the healthcheck itself, or just whether it's used in the calculation of UP or DOWN? Looks like it just turns them off, but the calculation thing might be useful?

Yep it does currently just outright turn them off. The logic being that you could enable/disable certain healthchecks programatically based on some kind of state/environment through a boolean

PR #306, opened 2024-12-09
Makes start of page markup match Plan History for spacing

pipwilson:
Before:

Screenshot 2024-12-09 at 14 45 24

After:

Screenshot 2024-12-09 at 15 19 53

### DIFF
--- a/server/views/pages/about.njk
+++ b/server/views/pages/about.njk
@@ -46,6 +46,10 @@
 
 {% block content %}
     {{ super() }}
+    {% if data.assessmentAreas.versionUpdatedAt %}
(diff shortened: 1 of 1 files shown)


server/views/pages/about.njk
```diff
@@ -46,6 +46,10 @@
 
 {% block content %}
     {{ super() }}
```
Thomas-Geraghty, line 48: I know this isn't from the PR, but any idea why this is here?

Thomas-Geraghty (approved):

server/views/pages/about.njk
pipwilson: I'll pick this up with you on slack.

PR #308, opened 2024-12-09
SP2-928 - Cypress test for about page

izzyL-DJ:
Cypress test for hard coded sentence information data in about page

### DIFF
--- a/integration_tests/e2e/about.cy.ts
+++ b/integration_tests/e2e/about.cy.ts
@@ -22,22 +22,10 @@ describe('Rendering', () => {
       })
   })
 
-  it('Should check if the hard-coded entries are displayed correctly', () => {
-    cy.get('tbody > :nth-child(2)')
-      .invoke('text')
-      .then(text => {
-        expect(text).to.include('Custodial Sentence  (4 years, 2 months and 6 days)')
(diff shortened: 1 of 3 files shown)


integration_tests/e2e/about.cy.ts
```diff
@@ -22,8 +22,22 @@ describe('Rendering', () => {
...
   })
 
-  it('Should check if the entries are displayed correctly', () => {
-    cy.visit('/about')
-    cy.get('.sentence-info').invoke('text').should('not.be.empty') // check table contents are not empty until we can check for specifics
+  it('Should check if the hard-coded entries are displayed correctly', () => {
+    cy.get('tbody > :nth-child(2)')
+      .invoke('text')
+      .then(text => {
+        expect(text).to.include('Custodial Sentence  (4 years, 2 months and 6 days)')
```
pipwilson: I'm not sure this is the right way of testing for these values because we will easily get false positives (e.g. the two tests for 'No' below will I think always succeed even if one of them is 'Yes') - I think it needs to check the individual cells. The cypress inspector will give you the relevant selector for each one - I suspect it will suggest something like `table.sentence-info td:nth-child(1)`.

pipwilson (approved):
LGTM

PR #310, opened 2024-12-10
Switch to using `eslint-config-hmpps` for linting rulesets

Thomas-Geraghty:
- Added `eslint-config-hmpps`
- Removed all existing eslint rulesets
- Added an override for `no-explicit-any` as we use this a few times in our code

**Note:** we should probably go clean up any `any` types


pipwilson (approved):
✍️

PR #315, opened 2024-12-10
SP2-928 about page cypresstests pt2

izzyL-DJ:


integration_tests/e2e/about.cy.ts
```diff
@@ -4,6 +4,9 @@ describe('Rendering', () => {
       cy.wrap(planDetails).as('plan')
       cy.openSentencePlan(planDetails.oasysAssessmentPk)
       cy.get('.moj-primary-navigation__container').contains('a', 'About').click()
+      cy.get(
+        ':nth-child(4) > .govuk-accordion__controls > .govuk-accordion__show-all > .govuk-accordion__show-all-text',
```
alexw-moj: How about `.govuk-accordion__show-all:first` instead?

integration_tests/e2e/about.cy.ts
```diff
@@ -28,4 +31,84 @@ describe('Rendering', () => {
...
+      'Accommodation',
+      'Personal relationships and community',
+      'Alcohol use',
+      'Employment and education',
+      'Drug use',
+      'Finances',
+      'Health and wellbeing',
+    ]
+    cy.get('.govuk-accordion__section-heading-text-focus')
+      .then($red => $red.toArray().map(el => el.innerText.trim())) // trim whitespace
```
alexw-moj: Why `$red`?

integration_tests/e2e/about.cy.ts
```diff
@@ -28,4 +31,84 @@ describe('Rendering', () => {
...
+      .should('deep.equal', expectedText)
+  })
+
+  it('Should check the hard-coded labels appear next to the correct, predetermined areas in correct order', () => {
+    cy.get('.govuk-accordion__section-summary-focus').eq(1).contains('Risk of reoffending')
```
alexw-moj: The numbers are a bit hard to follow, how about `cy.get('.govuk-accordion__section-heading').contains('Accommodation').contains('Risk of reoffending')` etc?

Might be worth also checking the others don't contain a badge too?

integration_tests/e2e/about.cy.ts
```diff
@@ -28,4 +31,84 @@ describe('Rendering', () => {
...
+    cy.get('.govuk-accordion__section-summary-focus').eq(4).contains('Risk of reoffending')
+  })
+
+  it('Should check all hard-coded links in each assessment section are in the expected order', () => {
+    cy.get('p.goal-link')
+      .eq(0)
+      .contains('Create thinking, behaviours and attitudes goal')
+      .should('have.attr', 'href')
+      .and('include', '/create-goal/thinking-behaviours-and-attitudes')
+    cy.get('p.goal-link')
```
alexw-moj: Similar for these with the numbers. How about this?
```
cy.contains('.govuk-accordion__section', 'Accommodation')
      .contains('a', 'Create accommodation goal')
      .should('have.attr', 'href')
      .and('include', '/create-goal/accommodation')
```

integration_tests/e2e/about.cy.ts
```diff
@@ -28,4 +31,84 @@ describe('Rendering', () => {
...
+      .should('have.attr', 'href')
+      .and('include', '/create-goal/finances')
+    cy.get('p.goal-link')
+      .eq(7)
+      .contains('Create health and wellbeing goal')
+      .should('have.attr', 'href')
+      .and('include', '/create-goal/health-and-wellbeing')
+  })
+
+  it('Should check if the hard-coded data for high-scoring areas are displayed correctly and in order', () => {
```
alexw-moj: Test name is a bit ambiguous, and only applies to one of the high scoring sections. More to come?
As above, numbers make the test a bit fragile.

integration_tests/e2e/about.cy.ts
izzyL-DJ: Short for reduced

.circleci/config.yml
```diff
@@ -35,7 +35,7 @@ jobs:
       - checkout
       - run:
           name: Update npm
-          command: 'sudo npm install -g npm@latest'
+          command: 'sudo npm install -g npm@next-10'
```
pipwilson: this can be changed back to "latest" now

Thomas-Geraghty:
I think this looks okay for a basic testing solution for the text content, but there's no real testing of the visual displays which is likely what our users wil rely on - considering they are usually quite time pressured. 

I think a separate PR that adds tests verifying how many `highscoring` `lowscoring` etc score blocks have been added to a score bar would be great

Thomas-Geraghty (approved):

PR #318, opened 2024-12-13
SP2-635 Flag Goals (for Now) without Steps

pipwilson:
This change adds a new validation model and runs all GETs to /plan through a new validation method, checking if the plan agreement status warrants running the validation model (i.e. if this is plan is not in Draft). 

The template logic for displaying error messages needs has had to be updated too, and the error summary now links directly to the correct goal summary card when it doesn't have steps.

Reviewers can start with server/routes/planOverview/PlanOverviewController.ts for the validation method and plan.njk for the templating


server/views/pages/plan.njk
```diff
@@ -92,26 +92,46 @@
 
 {% block content %}
     {{ super() }}
+
+    {% if data.plan.agreementStatus === 'AGREED' %}
+      {#                {% if data.plan.updatedDate > data.plan.agreementDate %}#TODO: SP2-836#}
+      <p class="govuk-body govuk-!-margin-top-5">{{ locale.lastUpdatedPlan }}</p>
+      {#                {% else %}#}
+      {#                    <p class="govuk-body">{{ locale.agreePlan }}</p>#}
+      {#                {% endif %}#}
```
Thomas-Geraghty, line 101: Leftover comments?

server/views/pages/plan.njk
```diff
@@ -92,26 +92,46 @@
...
                 }) %}
             {% endif %}
+
             {% for id, errorObj in errors.domain %}
                 {% if '.steps' in id %}
+                    {# id is of the form `goals.n.steps` and we extract which goal had the error #}
+                    {% set goalWithError = id[6] | int %}
+                    {% set goalLocale = interpolate(locale, {
+                       goalTitle: data.plan.goals[goalWithError].title
+                    }) %}
```
Thomas-Geraghty, line 126: I hadn't really thought about this being useable like so, neat

server/views/pages/plan.njk
```diff
@@ -54,7 +54,7 @@
 }) %}
 
 {% block pageTitle %}
-    {% if not errors.domain %}
+    {% if not (errors.domain | length) %}
```
Thomas-Geraghty, line 57: Any reason we don't just check the size of 'errors' in general? 

Is it because `errors` is always shaped like
```
{
   query: {},
   body: {},
   params: {},
   domain: {}
}
```
?

server/views/pages/plan.njk
pipwilson: This was already there, left in with the `#TODO: SP2-836`. In principle this is the logic for making 836 work in the template, but plan.updatedDate doesn't contain the right value. We thought we were going to get to that feature several months ago so it made sense to leave it in at the time.

server/views/pages/plan.njk
pipwilson: Yes, precisely. We can't check size or for null on the main `errors` object because it always has these fields.

Thomas-Geraghty: Ah, it'd be nice to fix that at some point!

Thomas-Geraghty (approved):

PR #319, opened 2024-12-16
Visual regression testing

pipwilson:
Adds the framework and a sample file for running visual regression testing.


cypress-image-diff-screenshots/baseline/visual-comparison.cy.ts-about-page.png
alexw-moj, line 1: There's more bits in the collapsed sections which we also want to compare?
Izzy found a nice `.click({ multiple: true })` for the accordion sections.

pipwilson: Yes, I think this is only the foundations - really the visual regression tests need integrating into the main tests in the same way as the accessibility ones I think.

package.json
```diff
@@ -137,6 +137,8 @@
     "cookie-session": "^2.1.0",
     "cypress": "^13.16.0",
     "cypress-axe": "^1.5.0",
+    "cypress-image-diff-html-report": "^2.2.0",
```
Thomas-Geraghty, line 141: What's the HTML report for? presumably just for seeing the results in the pipeline right?

pipwilson: Oh good question, I didn't actually set that up properly. Details are here: https://github.com/kien-ht/cypress-image-diff-html-report?tab=readme-ov-file#overview

PR #320, opened 2024-12-19
Updates to node 22

pipwilson:


tsconfig.json
```diff
@@ -1,17 +1,14 @@
...
-    "moduleResolution": "node",
     "outDir": "./dist",
     "sourceMap": true,
-    "skipLibCheck": true,
     "noEmit": false,
+    "strict": false,
     "allowJs": true,
     "checkJs": true,
     "resolveJsonModule": true,
-    "esModuleInterop": true,
```
Thomas-Geraghty, line 14: I'm guessing the removed properties appear in `@tsconfig/node22/tsconfig.json`?

Thomas-Geraghty (approved):

tsconfig.json
pipwilson: Yes. https://www.npmjs.com/package/@tsconfig/node22?activeTab=code

pipwilson: the new `strict` property is there to override the new default which is `true` without which our linting is very very sad.

PR #322, opened 2024-12-20
SP2-957 About Person page for Countersigning

pipwilson:
This PR:

1. Renames AboutPop to AboutPerson
2. Removes all the action buttons and links from the About Person page when a user has AccessMode.READ_ONLY
3. Adds Cypress tests for users accessing /about with READ_ONLY


integration_tests/e2e/about.cy.ts
```diff
@@ -10,8 +12,12 @@ describe('Rendering', () => {
   it('Should check the page rendered correctly', () => {
     cy.url().should('include', '/about')
     cy.get('h1').should('include.text', 'About')
-    cy.get('h2').contains('Sentence information')
-    cy.get('h2').contains('High-scoring areas from the assessment')
+    cy.get('h2').eq(0).contains('Sentence information')
+    cy.get('h2').eq(1).contains('High-scoring areas from the assessment')
```
Thomas-Geraghty, line 16: Using `eq` here to validate the titles are in the correct order right?

integration_tests/e2e/about.cy.ts
```diff
@@ -158,3 +151,29 @@ describe('Rendering', () => {
...
+  beforeEach(() => {
+    cy.createSentencePlan().then(planDetails => {
+      cy.wrap(planDetails).as('plan')
+      cy.openSentencePlan(planDetails.oasysAssessmentPk, AccessMode.READ_ONLY)
+      cy.get('.moj-primary-navigation__container').contains('a', 'About').click()
+    })
+  })
+
+  it('Should check the page rendered correctly with no Create Goal button', () => {
+    cy.url().should('include', '/about')
```
Thomas-Geraghty: Probably a throwaway thought, but I never know how to feel about verifying the URL is correct at the start of test if the URL was navigated to in the beforeEach step - why not just verify in the beforeEach?

integration_tests/e2e/about.cy.ts
```diff
@@ -158,3 +151,29 @@ describe('Rendering', () => {
...
+    cy.get('h1').should('include.text', 'About')
+    cy.get('h2').eq(0).contains('Sentence information')
+    cy.get('h2').eq(1).contains('High-scoring areas from the assessment')
+    cy.get('[role="button"]').should('have.length', 1)
```
Thomas-Geraghty, line 168: No change needed but, just interested in your opinion;
* This will break the test if we ever added a clickable button elsewhere into the About Person page that is useable in the READ_ONLY view.
* Counterpoint, if the page has been updated, the test should likely be updated anyway

server/views/pages/about.njk
```diff
@@ -36,7 +36,7 @@
             href: "/create-goal/accommodation",
             classes: "govuk-button--secondary",
             type: 'button'
-        }
+        } if data.readWrite === true
```
Thomas-Geraghty, line 39: is the `=== true` needed in this case?

integration_tests/e2e/about.cy.ts
pipwilson: Yes.

integration_tests/e2e/about.cy.ts
pipwilson: Yes, this has probably been propagated via copy/paste over time. I'll take this out from here.

integration_tests/e2e/about.cy.ts
pipwilson: Yes I think I would want this test to break if another button was added because as you say, the selector for this particular button will need to be updated.

server/views/pages/about.njk
pipwilson: No, but I quite like super-explicit checks for permission-based validation - I can take it out if you prefer.

integration_tests/e2e/about.cy.ts
pipwilson: Fixed in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/322/commits/3dfc627af170843f68f4b609ff6f0d4803478718

Thomas-Geraghty (approved):

PR #332, opened 2025-01-08
SP2-1110: Fixed typo causing additional info to not show on Finances section

JakeGoodwin22:
Screenshot 2025-01-08 at 10 37 35


pipwilson (approved):
LGTM! 🚀

PR #334, opened 2025-01-09
SP2-1111: Additional details not showing for 'No' answers

JakeGoodwin22:
Screenshot 2025-01-08 at 16 10 36


server/views/components/assessment/_assessment-area.njk
```diff
@@ -29,20 +29,26 @@
...
                         <h3 class="govuk-heading-s">{{ locale.detail.notLinkedToRoSH }}</h3>
                     {% endif %}
 
-                    {% if linkedtoReoffending or riskOfReoffendingData %}
+                    {% if linkedtoReoffending %}
                         <div class="ror">
                             <h3 class="govuk-heading-s">{{ locale.detail.linkedToReoffending }}</h3>
                             <p class="sp-text-pre-wrap govuk-body">{{ riskOfReoffendingData }}</p>
                         </div>
+                    {% elif riskOfReoffendingData %}
```
pipwilson: above you use elseif - I know they're the same but we should prefer one or the other in the same file I think.

JakeGoodwin22: Good spot! I've changed it to 'elseif'

server/utils/assessmentUtils.ts
```diff
@@ -33,13 +33,17 @@ export const formatAssessmentData = (
...
         assessment.sanAssessmentData[
-          `${areaConfig.assessmentKey}_practitioner_analysis_risk_of_serious_harm_yes_details`
+          `${areaConfig.assessmentKey}_practitioner_analysis_risk_of_serious_harm_${linkedToRoSHYesNo}_details`
         ]?.value
       const riskOfReoffendingDetails =
         assessment.sanAssessmentData[
-          `${areaConfig.assessmentKey}_practitioner_analysis_risk_of_reoffending_yes_details`
+          `${areaConfig.assessmentKey}_practitioner_analysis_risk_of_reoffending_${linkedToReoffendingYesNo}_details`
         ]?.value
       const strengthsOrProtectiveFactors =
```
pipwilson: do we need to do something similar for this as well? can there be details when the ...strengths_or_protective_factors... is no?

server/@types/Assessment.ts
```diff
@@ -15,6 +15,7 @@ export interface AssessmentArea {
   riskOfReoffendingDetails?: string
   linkedtoReoffending: boolean
   motivationToMakeChanges?: string
+  linkedtoStrengthsOrProtectiveFactors: boolean
```
pipwilson: so is this an actual field that comes in from SAN?

server/utils/assessmentUtils.ts
```diff
@@ -28,6 +28,8 @@ export const formatAssessmentData = (
       const linkedtoRoSH = crimNeeds[areaConfig.crimNeedsKey][`${areaConfig.crimNeedsSubKey}LinkedToHarm`] === 'YES'
       const linkedtoReoffending =
         crimNeeds[areaConfig.crimNeedsKey][`${areaConfig.crimNeedsSubKey}LinkedToReoffending`] === 'YES'
+      const linkedtoStrengthsOrProtectiveFactors =
```
pipwilson: given that we also have the YesNo field below, would it make sense to rename this (and the two above it) to `isLinkedToStrengthsOrProtectiveFactors` to try and make it a bit clearer?

pipwilson: e.g. I'm slightly surprised that something like this is not the default for eslint https://typescript-eslint.io/rules/naming-convention/#enforce-that-boolean-variables-are-prefixed-with-an-allowed-verb

server/views/components/assessment/_assessment-area.njk
```diff
@@ -7,6 +7,8 @@
         {% set linkedtoReoffending = assessment.linkedtoReoffending %}
         {% set riskOfSeriousHarmData = assessment.riskOfSeriousHarmDetails %}
         {% set riskOfReoffendingData = assessment.riskOfReoffendingDetails %}
+        {% set linkedtoStrengthsOrProtectiveFactorsData = assessment.linkedtoStrengthsOrProtectiveFactorsData %}
```
pipwilson: you've added this to make it consistent, but I wonder if we need these fields at all now that you've removed the `or` statement from the `if` lines. The other `if`s all address via the assessment object e.g. `{% if assessment.motivationToMakeChanges %}` .

pipwilson (approved):
🚀

PR #340, opened 2025-01-13
SP2-1032: Go back from Add or change steps page

JakeGoodwin22:
Note: once we go back to the change goal page, using the back button. I would expect that we could press submit again and go once more to the add steps page, so have implemented this to get others opinions on it. Before making this change, clicking submit on the change-goal page would take us back to the plan overview, which doesn't seem right

### DIFF
--- a/server/routes/add-steps/AddStepsController.ts
+++ b/server/routes/add-steps/AddStepsController.ts
@@ -59,8 +59,10 @@ export default class AddStepsController {
     try {
       await req.services.stepService.saveAllSteps(goalData, req.params.uuid)
 
-      const link = `${URLs.PLAN_OVERVIEW}?type=current`
-      req.services.sessionService.setReturnLink(null)
+      const link =
+        req.services.sessionService.getReturnLink() === `/change-goal/${req.params.uuid}/`
+          ? `${URLs.PLAN_OVERVIEW}?type=current`
+          : req.services.sessionService.getReturnLink()
 
       return res.redirect(link)
     } catch (e) {
--- a/server/routes/add-steps/AddStepsController.ts
+++ b/server/routes/add-steps/AddStepsController.ts
@@ -59,8 +59,7 @@ export default class AddStepsController {
     try {
       await req.services.stepService.saveAllSteps(goalData, req.params.uuid)
 
-      // TODO: this is wrong... we should be able to add steps only, or add steps after adding a (current or future) goal, and get a success banner...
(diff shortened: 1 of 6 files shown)


server/routes/changeGoal/ChangeGoalController.ts
```diff
@@ -61,6 +61,11 @@ export default class ChangeGoalController {
       const planUuid = req.services.sessionService.getPlanUUID()
       const plan = await req.services.planService.getPlanByUuid(planUuid)
 
+      // Check if the user has came back from the add-goals page
+      if (req.services.sessionService.getReturnLink() === `/change-goal/${goalUuid}/`) {
+        req.services.sessionService.setReturnLink(null)
+        redirectTarget = `/goal/${goalUuid}/add-steps`
```
Thomas-Geraghty: Do we not want to keep something set in the returnLink variable? Could this maybe cause a null exception somewhere? or a user being redirected to `null`?

server/routes/add-steps/AddStepsController.ts
```diff
@@ -59,8 +59,7 @@ export default class AddStepsController {
     try {
       await req.services.stepService.saveAllSteps(goalData, req.params.uuid)
 
-      // TODO: this is wrong... we should be able to add steps only, or add steps after adding a (current or future) goal, and get a success banner...
-      const link = req.services.sessionService.getReturnLink() ?? `${URLs.PLAN_OVERVIEW}?status=success`
+      const link = `${URLs.PLAN_OVERVIEW}?type=current`
       req.services.sessionService.setReturnLink(null)
```
Thomas-Geraghty: Same concern here as https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/340/files#r1916598230

server/routes/changeGoal/ChangeGoalController.ts
JakeGoodwin22: Yep, you're right. I can't actually remember why I added this line in. I'll remove it so it remains set to the current returnLink (which solves a problem on the next page in the journey)

server/routes/add-steps/AddStepsController.ts
JakeGoodwin22: Yep, no idea what I was thinking with adding these in. I have changed it so that if the returnLink is pointing to change goal, then we will redirect to the plan overview (as it means they are adding a new goal) otherwise we use the return link, this means that if they are changing any steps, such as when it is an agreed plan it will go back to the update goals and steps page.

Thomas-Geraghty (approved):

PR #345, opened 2025-01-15
SP2-598 re-add goal to plan

pipwilson:
This PR adds support for re-adding a goal back to a plan after it has been removed.

The end-to-end tests will fail until https://github.com/ministryofjustice/hmpps-sentence-plan/pull/164 is merged.

This also adds backend API support for agreeing a plan and removing a goal so that we don't have to go through those pages when they are secondary to the main purpose of the tests.

I also extracted some duplicated functionality from several controllers and put it in a new utils class used when setting goal target dates.

### DIFF
--- a/server/services/sentence-plan/goalService.ts
+++ b/server/services/sentence-plan/goalService.ts
@@ -17,6 +17,7 @@ export default class GoalService {
     return restClient.put<Goal>({ path: `/goals/${goalUuid}`, data: goal })
   }
 
+  // rename this to updateGoalStatus
   async updateGoal(goal: Partial<NewGoal>, goalUuid: string) {
     const restClient = await this.sentencePlanApiClient.restClient('Update goal data')
     return restClient.patch<Goal>({ path: `/goals/${goalUuid}`, data: goal })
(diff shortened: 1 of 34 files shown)


server/services/sentence-plan/goalService.ts
```diff
@@ -17,6 +17,7 @@ export default class GoalService {
     return restClient.put<Goal>({ path: `/goals/${goalUuid}`, data: goal })
   }
 
+  // rename this to updateGoalStatus
```
JakeGoodwin22: Does this need renaming?

pipwilson: Yes. I should have added a TODO note! I'll do this.

JakeGoodwin22 (dismissed):
Noted some stuff still needs updating but overall looks good, only some very minor changes such as updating the function name.

server/services/sentence-plan/goalService.ts
pipwilson: done in 7c00fc4719405bbc06035ef65f855887cd24ab34

JakeGoodwin22 (approved):

PR #349, opened 2025-01-16
Back link selects correct tab based on goal type

alexw-moj:
Following design review it was noticed that if a future goal was selected for update that the back link did not select the right link on the plan page.

### DIFF
--- a/server/routes/update-goal/UpdateGoalController.ts
+++ b/server/routes/update-goal/UpdateGoalController.ts
@@ -30,7 +30,7 @@ export default class UpdateGoalController {
       const mainAreaOfNeed = sortedAreasOfNeed.find(areaOfNeed => areaOfNeed.name === goal.areaOfNeed.name)
       const relatedAreasOfNeed = goal.relatedAreasOfNeed.map(need => need.name)
 
-      const returnLink = req.services.sessionService.getReturnLink()
+      const returnLink = `/plan?type=${goalType}`
       req.services.sessionService.setReturnLink(`/update-goal-steps/${uuid}`)
 
       return res.render('pages/update-goal', {
@@ -42,7 +42,6 @@ export default class UpdateGoalController {
           mainAreaOfNeed,
           relatedAreasOfNeed,
           returnLink,
-          goalType,
         },
         errors,
       })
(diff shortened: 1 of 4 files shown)


server/views/pages/update-goal.njk
```diff
@@ -53,7 +53,7 @@
 
 {% block content %}
     {{ super() }}
-    <a href="/plan" class="govuk-back-link">{{ locale.common.backLink.text }}</a>
+    <a href="/plan?type={{ data.goalType }}" class="govuk-back-link">{{ locale.common.backLink.text }}</a>
```
pipwilson: could we use data.returnLink for the whole href like the other templates please? or is there a reason for not doing that?

alexw-moj: If you go to `Update > Add or change steps` then that will break the back link if I use `data.returnLink`.

pipwilson: I don't think this is right. `returnLink` is not set to that value. That value is put in the session. `returnLink` is then passed to the template but not used.

pipwilson: hang on, your comment changed when I was writing mine - will re-review :)

server/routes/update-goal/UpdateGoalController.ts
```diff
@@ -25,11 +25,12 @@ export default class UpdateGoalController {
     try {
       const sortedAreasOfNeed = this.referentialDataService.getSortedAreasOfNeed()
       const goal = await req.services.goalService.getGoal(uuid)
+      const goalType: string = goalStatusToTabName(goal.status)
       const popData = req.services.sessionService.getSubjectDetails()
       const mainAreaOfNeed = sortedAreasOfNeed.find(areaOfNeed => areaOfNeed.name === goal.areaOfNeed.name)
       const relatedAreasOfNeed = goal.relatedAreasOfNeed.map(need => need.name)
 
-      const returnLink = req.services.sessionService.getReturnLink()
+      const returnLink = `/plan?type=${goalType}`
```
pipwilson, line 33: I think this looks better. Just checking: is this always true regardless of previous page?

alexw-moj: There's only one way into this page, from the plan page. Also checked with Design after this comment :)

pipwilson (approved):
yeah!

PR #352, opened 2025-01-20
SP2-1095 Remove Mark as Achieved from Plan page

alexw-moj:
Removes the mark as achieved link from the actions on the summary card, moving it to the Update Goal page.

### DIFF
--- a/server/routes/update-goal/UpdateGoalController.ts
+++ b/server/routes/update-goal/UpdateGoalController.ts
@@ -87,6 +87,10 @@ export default class UpdateGoalController {
 
       await this.updateSteps(req, goal, steps, note)
 
+      if (req.body.action === 'mark-as-achieved') {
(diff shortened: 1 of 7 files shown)


server/routes/update-goal/UpdateGoalController.ts
```diff
@@ -87,6 +87,10 @@ export default class UpdateGoalController {
 
       await this.updateSteps(req, goal, steps, note)
 
+      if (req.body.action === 'mark-as-achieved') {
```
pipwilson, line 90: we should probably add a unit test for this.

alexw-moj: Added unit test in: https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/352/commits/9f3a76c98d6524979a485431da5ab91fff4c2a86

pipwilson (approved):
🚀 🚀 🚀

PR #354, opened 2025-01-20
SP2-1223: Null answers - show as not answered

JakeGoodwin22:
Screenshot 2025-01-23 at 12 04 39


server/views/pages/about-not-complete.njk
```diff
@@ -0,0 +1,107 @@
...
+            {%  for error in errors.domain %}
+                {%  set errorList = (errorList.push({ text: locale.errors[error]}), errorList)    %}
+            {%  endfor %}
+                {{ govukErrorSummary({
+                    titleText: "There is a problem",
+                    errorList: errorList
+                }) }}
+            {% endif %}
+
+            {% if data.deliusData.sentences.length > 0 %}
```
pipwilson, line 68: is this identical to the section in about.njk? if so, can we extract it and include it instead of duplicating?

server/views/pages/about-not-complete.njk
```diff
@@ -0,0 +1,107 @@
...
+                            {% else %}
+                                No
+                            {% endif %}
+                        </td>
+                    </tr>
+                {% endfor  %}
+            </table>
+            {% endif %}
+
+            <div class="govuk-warning-text">
```
pipwilson: Is there a reason you're not using the macro from the design system?

server/routes/aboutPerson/AboutPersonController.ts
```diff
@@ -32,6 +32,24 @@ export default class AboutPersonController {
       const pageId = 'about'
       const readWrite = req.services.sessionService.getAccessMode() === AccessMode.READ_WRITE
 
+      const areas = [...assessmentAreas.lowScoring, ...assessmentAreas.highScoring, ...assessmentAreas.other]
+
+      for (const area of areas) {
```
pipwilson, line 37: I think instead of duplicating the bulk of the return statement you could do something like `let returnPage = 'pages/about'; if(areas.some(area => area.isMissingInformation)) { returnPage = 'pages/about-not-complete' }` and then just do res.render(returnPage, ...

server/routes/aboutPerson/locale.json
```diff
@@ -14,9 +14,9 @@
       "lastUpdated": "Last updated on {{ lastUpdatedDate }}.",
       "RoSH": "RoSH (Risk of Serious Harm)",
       "riskOfReoffending": "Risk of reoffending",
-      "linkedToRoSH": "This area is linked to RoSH (risk of serious harm)",
-      "notLinkedToRoSH": "This area is not linked to RoSH (risk of serious harm)",
-      "linkedToReoffending": "This area is linked to risk of reoffending",
+      "linkedtoHarm": "This area is linked to RoSH (risk of serious harm)",
```
pipwilson: change to 'Harm' is OK but can we keep the camel case on the rest of the field names?

server/utils/assessmentUtils.ts
```diff
@@ -25,35 +25,49 @@ export const formatAssessmentData = (
...
 
-      const linkedToRoSHYesNo = isLinkedtoRoSH ? 'yes' : 'no'
-      const linkedToReoffendingYesNo = isLinkedtoReoffending ? 'yes' : 'no'
-      const linkedtoStrengthsOrProtectiveFactorsYesNo = isLinkedtoStrengthsOrProtectiveFactors ? 'yes' : 'no'
-
       const riskOfSeriousHarmDetails =
-        assessment.sanAssessmentData[
-          `${areaConfig.assessmentKey}_practitioner_analysis_risk_of_serious_harm_${linkedToRoSHYesNo}_details`
-        ]?.value
+        linkedtoHarm !== 'null'
```
pipwilson: is 'null' a possible value here? If the field is missing in line 29, what does calling "toLowerCase" on it do? I'm guessing it throws an exception so we won't get to this line.

server/utils/assessmentUtils.ts
```diff
@@ -25,35 +25,49 @@ export const formatAssessmentData = (
...
+          ? assessment.sanAssessmentData[
+              `${areaConfig.assessmentKey}_practitioner_analysis_risk_of_serious_harm_${linkedtoHarm}_details`
```
pipwilson: presumably this kind of null check is what we should be doing up on lines 29-32 now that we're not generating the boolean values

server/utils/assessmentUtils.ts
```diff
@@ -25,35 +25,49 @@ export const formatAssessmentData = (
...
-        ]?.value
+        linkedtoStrengthsOrProtectiveFactors !== 'null'
+          ? assessment.sanAssessmentData[
+              `${areaConfig.assessmentKey}_practitioner_analysis_strengths_or_protective_factors_${linkedtoStrengthsOrProtectiveFactors}_details`
+            ]?.value
+          : undefined
 
       score = crimNeeds[areaConfig.crimNeedsKey][`${areaConfig.crimNeedsSubKey}OtherWeightedScore`]
+
+      const isMissingInformation =
```
pipwilson, line 65: if we can make these all either null or undefined (by making the code above which sets the values consistent) then you can do something like `isMissingInformation = [linkedToHarm, LinkedToReoffend, etc].every(field => field !== null)`

server/@types/Assessment.ts
```diff
@@ -11,11 +11,12 @@ export interface AssessmentArea {
   title: string
   overallScore?: string
   riskOfSeriousHarmDetails?: string
-  isLinkedtoRoSH: boolean
+  linkedtoHarm: string
```
pipwilson: Ah, I see "to" was already not camelcase - it should be! Can you rename please?

server/views/pages/about-not-complete.njk
JakeGoodwin22: I left it like this because the plan is to remove this new page anyway, now the new designs have been done. It prevented more clean up later.

server/views/pages/about-not-complete.njk
JakeGoodwin22: Updated now, just done that way as it was a nice easy copy and paste from the GDS for me to demo at the time. Good spot, needed changing. I had forgot

server/routes/aboutPerson/AboutPersonController.ts
JakeGoodwin22: Done!

server/routes/aboutPerson/locale.json
JakeGoodwin22: Done!

server/utils/assessmentUtils.ts
JakeGoodwin22: Discussed this on slack, adding here for visibility. The value is sent to us as "NULL" rather than being undefined, so it is a string. We do .toLowerCase() so we can call the correct key for each field such as ..._practitioner_analysis_risk_of_serious_harm__**yes**_details. Added comment in the code to make this clearer in the future

server/views/pages/about-not-complete.njk
pipwilson: OK - can you add a subtask to the Jira story which has the "real" solution on it to delete this page please?

server/views/pages/about-not-complete.njk
```diff
@@ -91,15 +98,11 @@
...
 
-            <div class="govuk-warning-text">
-            <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
-              <strong class="govuk-warning-text__text">
-                <span class="govuk-visually-hidden">Warning</span>
-                  This information is not available yet because the assessment has not been completed. You can <a href="{{data.oasysReturnUrl}}">check the summary screen in OASys</a> instead.
-              </strong>
-            </div>
+            {{ govukWarningText({
+              html: '<strong>This information is not available yet because the assessment has not been completed. You can <a href="{{ data.oasysReturnUrl }}">check the summary screen in OASys</a> instead.</strong>',
```
pipwilson: should probably be in the locale file

server/views/pages/about-not-complete.njk
JakeGoodwin22: I'll sort this out 👍

pipwilson (approved):
LGTM

PR #356, opened 2025-01-21
Update base docker image from bullseye to bookworm

pipwilson:
This PR moves us to the same Docker image as the HMPPS TS template application (https://github.com/ministryofjustice/hmpps-template-typescript/blob/main/Dockerfile).

It **may** resolve two HIGH security vulnerabilities, but running trivy locally does not confirm or refute this :/


Dockerfile
```diff
@@ -27,10 +27,10 @@ ENV GIT_BRANCH=${GIT_BRANCH}
 RUN apt-get update && \
     apt-get upgrade -y && \
     apt-get autoremove -y && \
-    apt-get install -y make python g++ curl && \
+    apt-get install -y make python3 g++ curl && \
```
alexw-moj: We don't need Python installed anymore, I think.

pipwilson: Removed in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/356/commits/e6dc3a148b96af4be9bb09838187b360a8a888f1

alexw-moj (approved):

PR #370, opened 2025-01-27
SP2-1222: Update crest of header component

aiden-herbert:
Update to crown crest image in header.

Before update:
Screenshot 2025-01-27 at 11 16 57

After update:
Screenshot 2025-01-27 at 11 12 09


izzyL-DJ (dismissed):
LGTM

server/views/partials/header.njk
```diff
@@ -4,9 +4,33 @@
...
 
-          <image src="/assets/images/crest.svg" xlink:href="" width="40" height="40"></image>
+          <image src="/assets/images/moj-logotype-crest-2024.png" xlink:href="" width="40" height="40"></image>
```
alexw-moj, line 33: What does this refer to?

alexw-moj, line 33: This is in MOJ Frontend, but does the build copy it into our assets?

aiden-herbert: No, this solution follows a similar approach done for SAN however, which is a mapped path to the asset in question. From what I can tell this limits overhead and is referenced in one place to be used by other projects that are changing the crest statically without the component (as we are here).

aiden-herbert: That said @alexw-moj you prompted me to look in the assets/images folder and I have spotted in the assets the old image that can be removed. I shall do that in the next commit :-).

- [x] Remove image assets that are no longer required.

izzyL-DJ (approved):
LGTM

alexw-moj:
Maybe leave the `favicon.ico`? If you 'View Source' it's still referenced in the header via `node_modules/govuk-frontend/dist/govuk/template.njk`.

Thomas-Geraghty:
> Maybe leave the favicon.ico? If you 'View Source' it's still referenced in the header via node_modules/govuk-frontend/dist/govuk/template.njk.

It's actually `favicon.svg` that's referenced, which seems to be an asset served from the govuk-frontend assets folder

PR #372, opened 2025-01-28
SP2-1223: Removed optional question from assessment complete check

JakeGoodwin22:
- MotivationToMakeChanges is not needed as it is an optional question, therefore is not always present.
- Removed additional logging no longer required


pipwilson (approved):
cool

PR #374, opened 2025-01-31
SP2-633: Confirm a goal is achieved

JakeGoodwin22:
Screenshot 2025-01-31 at 09 25 20


server/routes/achieveGoal/locale.json
```diff
@@ -1,21 +1,27 @@
...
+      "is-goal-achieved-radio": {
+        "isNotEmpty": "Select if they have achieved this goal"
+      },
       "goal-achievement-helped": {
         "maxLength": "How achieving this goal has helped must be 4,000 characters or less"
       }
     },
-    "confirmGoalAchievedButton": "Confirm",
-    "doNotMarkAsAchievedLink": "Do not mark as achieved"
+    "confirmGoalAchievedButton": "Save and continue"
```
pipwilson, line 25: The screenshots in Jira say "Save" but the table of changes say "Save and continue" - which is correct?

integration_tests/pages/plan-overview.ts
```diff
@@ -31,4 +31,8 @@ export default class PlanOverview {
   clickDownOnSummaryCard = (value: number) => {
     this.getSummaryCard(value).contains('.govuk-button', 'Move goal down').click()
   }
+
+  isGoalAchievedRadio = (value: string) => {
```
pipwilson: presumably this should be in a new remove-goal.ts file rather than plan-overview?

server/routes/achieveGoal/locale.json
JakeGoodwin22: Confirmed with Lou "Save an continue" is correct

integration_tests/pages/plan-overview.ts
JakeGoodwin22: Agreed, it definitely should. Moved it now 👍

pipwilson: oops, i was looking at Aiden's remove goal work when reviewing this, I did indeed mean achieve goal!

pipwilson (approved):
✅

PR #375, opened 2025-01-31
SP2 1228: Remove Goal Title

aiden-herbert:
This work had already been done prior by @pipwilson, so the wording requested for the ticket was already in place. There were some improvements and additions Cypress wise that could be done, which has been addressed in this PR to verify UI content and ensure tests are testing where they should be.


JakeGoodwin22 (approved):
LGTM 👍

PR #376, opened 2025-02-03
Fix no date showing after marking as achieved

JakeGoodwin22:
On the Achieved goals tab "Marked as achieved on [Date]" was not displaying

### DIFF
--- a/integration_tests/e2e/achieve-goal.cy.ts
+++ b/integration_tests/e2e/achieve-goal.cy.ts
@@ -87,7 +87,7 @@ describe('Achieve goal', () => {
       cy.url().should('include', 'plan?type=achieved&status=achieved')
       cy.get(':nth-child(3) > .moj-sub-navigation__link')
       cy.get('.moj-sub-navigation__link').eq(2).should('contain', 'Achieved goals (1)')
-      cy.get('.govuk-body').should('contain', 'Marked as achieved on')
+      cy.get('.govuk-summary-card').should('contain', 'Marked as achieved on')
 
       cy.get<Goal>('@newGoal').then(goal => {
         cy.visit(`/view-achieved-goal/${goal.uuid}`)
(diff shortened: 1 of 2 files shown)


server/views/components/summary-card/goal-summary-card.njk
```diff
@@ -39,6 +39,8 @@
             <div class="goal-date-and-notes">
             {% if params.goal.targetDate and params.goal.status != "ACHIEVED" and params.goal.status != "REMOVED" %}
                 <p>{{ locale.info.aimToAchieveDate }}</p>
+            {% elif params.goal.status == "ACHIEVED" %}
+                <p>{{ locale.info.achievedDate }}</p>
```
aiden-herbert, line 43: General question: Is there a change to the UI at all following the logic being moved and changed to an `else if` over an `if` statement?

JakeGoodwin22: The change is more to maintain consistency with our other logic than anything else. When it was an 'if' it suggested that you can have multiple status' so I changed it so it's clear that we only have one at any time per goal summary card

integration_tests/e2e/achieve-goal.cy.ts
```diff
@@ -87,6 +87,7 @@ describe('Achieve goal', () => {
       cy.url().should('include', 'plan?type=achieved&status=achieved')
       cy.get(':nth-child(3) > .moj-sub-navigation__link')
       cy.get('.moj-sub-navigation__link').eq(2).should('contain', 'Achieved goals (1)')
+      cy.get('.govuk-body').should('contain', 'Marked as achieved on')
```
alexw-moj: Can you make the selector a bit more specific to the card, please?

JakeGoodwin22: Yep, done

alexw-moj (approved):

PR #377, opened 2025-02-03
SP2-1226 Go back to Delete Confirmation page from Change Goal page

alexw-moj:
Updating the Back button behaviour of the Change goal page if we have come to that page via a link on the Delete goal page.


JakeGoodwin22 (approved):
LGTM

PR #378, opened 2025-02-03
SP2-633: Fix for 'Mark as achieved' button

JakeGoodwin22:
- reverts changes that were removed by the original change, so 'Mark as achieved' goes to the original page
- Adds new confirm-achieve-goal for new functionality


integration_tests/e2e/confirm-achieve-goal.cy.ts
```diff
@@ -0,0 +1,152 @@
...
+
+          cy.get('#step-status-1').select('Completed') // select completed status
+
+          cy.get('button').contains('Save goal and steps').click()
+        })
+      })
+    })
+
+    it('Confirm goal achieved successfully without optional note', () => {
+      achieveGoal.isGoalAchievedRadio('yes')
```
pipwilson, line 90: can we have a validation here to confirm the URL you're on please? Just to help with following what's going on.

integration_tests/e2e/confirm-achieve-goal.cy.ts
```diff
@@ -0,0 +1,152 @@
...
+      cy.get('button').contains('Save and continue').click()
+      cy.url().should('include', 'plan?type=achieved&status=achieved')
+      cy.get(':nth-child(3) > .moj-sub-navigation__link')
```
pipwilson: what does this line do?

integration_tests/e2e/confirm-achieve-goal.cy.ts
```diff
@@ -0,0 +1,152 @@
...
+      cy.get('.moj-sub-navigation__link').eq(2).should('contain', 'Achieved goals (1)')
+      cy.get('.govuk-summary-card').should('contain', 'Marked as achieved on')
+
+      cy.get<Goal>('@newGoal').then(goal => {
```
pipwilson: can't we just click the "view details" link in the summary card? probably follows the above a bit more cleanly.

integration_tests/e2e/confirm-achieve-goal.cy.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    it('Confirm goal achieved successfully with optional note', () => {
+      achieveGoal.isGoalAchievedRadio('yes')
+      cy.get('#goal-achievement-helped').type('Some optional text in the achievement note field')
+      cy.get('button').contains('Save and continue').click()
+      cy.url().should('include', 'plan?type=achieved&status=achieved')
+      cy.get(':nth-child(3) > .moj-sub-navigation__link')
+      cy.get('.moj-sub-navigation__link').eq(2).should('contain', 'Achieved goals (1)')
+
+      cy.get('.moj-sub-navigation__link').eq(2).click()
```
pipwilson: this section of this test and the one above don't do quite the same things - can they be made the same please?

integration_tests/e2e/confirm-achieve-goal.cy.ts
```diff
@@ -0,0 +1,152 @@
...
+      planOverview.getSummaryCard(0).get('a').contains('View details').click()
+      cy.get('.govuk-details__text').should('contain', 'Some optional text in the achievement note field')
+      cy.get('p.govuk-body').should('contain', 'Marked as achieved on ')
+      cy.get('a.govuk-back-link').should('have.attr', 'href').and('include', '/plan?type=achieved')
+      cy.get('.govuk-button--secondary').should('have.length', 0)
+      cy.checkAccessibility()
+    })
+
+    it('Confirm errors are displayed with optional note of more than 4000 characters', () => {
+      const lorem = faker.lorem.paragraphs(40)
```
pipwilson: over in add-steps.cy.ts we have a function called `generateStringOfLength` - would we be better off extracting that into a utils class we can call in cases like this? It would be useful to be able to test with a string of a known length I think.

integration_tests/e2e/confirm-achieve-goal.cy.ts
```diff
@@ -0,0 +1,152 @@
...
+      cy.url().should('include', '/confirm-if-achieved/')
+      cy.get('.govuk-error-summary').should(
+        'contain',
+        'How achieving this goal has helped must be 4,000 characters or less',
+      )
+      cy.get('#goal-achievement-helped-error').should(
+        'contain',
+        'How achieving this goal has helped must be 4,000 characters or less',
+      )
+      cy.get('#goal-achievement-helped').should('not.be.empty')
```
pipwilson: isn't this always true if line 122 has passed?

server/routes/confirmAchieveGoal/ConfirmAchieveGoalController.ts
```diff
@@ -0,0 +1,87 @@
...
+import validateRequest from '../../middleware/validationMiddleware'
+import transformRequest from '../../middleware/transformMiddleware'
+import ConfirmAchieveGoalPostModel from './models/ConfirmAchieveGoalPostModel'
+import { requireAccessMode } from '../../middleware/authorisationMiddleware'
+import { AccessMode } from '../../@types/Handover'
+import { HttpError } from '../../utils/HttpError'
+import URLs from '../URLs'
+import { goalStatusToTabName } from '../../utils/utils'
+
+export default class ConfirmAchieveGoalController {
```
pipwilson, line 15: because I am simple, can we have a comment at the top of this and the AchieveGoalController please with what each is for? I'm sure we will forget in 6 months' time :)

integration_tests/e2e/confirm-achieve-goal.cy.ts
JakeGoodwin22: Yes, added

integration_tests/e2e/confirm-achieve-goal.cy.ts
JakeGoodwin22: Nothing by the looks of it, it was copied directly from the original test class. I have removed this line in both classes

integration_tests/e2e/confirm-achieve-goal.cy.ts
JakeGoodwin22: Yeah, makes sense to do it that way

integration_tests/e2e/confirm-achieve-goal.cy.ts
JakeGoodwin22: Updated 👍

integration_tests/e2e/confirm-achieve-goal.cy.ts
JakeGoodwin22: added integrationUtils, yeah it's definitely useful and easier

integration_tests/e2e/confirm-achieve-goal.cy.ts
JakeGoodwin22: This is checking that the data in the textbox still exists when we have a validation error. To make this a bit clearer I have changed it to check against the content rather than just whether it is not empty

server/routes/confirmAchieveGoal/ConfirmAchieveGoalController.ts
JakeGoodwin22: Yep, not a problem

pipwilson (approved):
🚀 LGTM 🚀

PR #379, opened 2025-02-04
SP2-1227: Hide plan-history for draft plan

JakeGoodwin22:


server/routes/plan-history/PlanHistoryController.ts
```diff
@@ -13,6 +13,10 @@ export default class PlanHistoryController {
       const notes = await req.services.planService.getNotes(planUuid)
       const readWrite = req.services.sessionService.getAccessMode() === AccessMode.READ_WRITE
 
+      if (notes.length === 0) {
+        return next(HttpError(403, 'Plan has not been agreed'))
```
Thomas-Geraghty, line 17: Might be worth knowing that the error message is actually only displayed in development builds (specifically as part of the stack trace print).

server/routes/planOverview/PlanOverviewController.ts
```diff
@@ -49,6 +49,7 @@ export default class PlanOverviewController {
       return res.render(page, {
         locale: locale.en,
         data: {
+          planAgreementStatus: this.plan.agreementStatus, // required by layout.njk
```
Thomas-Geraghty, line 52: As mentioned on Slack, considering this is now being required at the base level for all templates, it's likely worth us exploring a better means to provide this data to views rather than requiring it in every controller. One for a follow-up ticket.

Thomas-Geraghty (approved):

PR #380, opened 2025-02-04
SP2-1263 Updates status name to "changed" rather than "updated" and adds tests

pipwilson:
We had not updated the status field name in the ChangeGoalController from "updated" to "changed". This PR does that and adds additional tests to make sure we really are on the correct tab.


JakeGoodwin22 (approved):
LGTM!

PR #382, opened 2025-02-07
SP2-948 Plan Overview (Post-Agree) - Amendments

izzyL-DJ:
Screenshot 2025-02-07 at 12 32 00


server/views/components/summary-card/goal-summary-card.njk
```diff
@@ -56,36 +57,52 @@
...
+                                    <strong class="govuk-tag govuk-tag--grey">{{ locale.stepsList.status.notStarted }}</strong>
+                                {% elseif step.status == 'IN_PROGRESS' %}
+                                    <strong class="govuk-tag govuk-tag">{{ locale.stepsList.status.inProgress }}</strong>
+                                {% elseif step.status == 'COMPLETED' %}
+                                    <strong class="govuk-tag govuk-tag--green">{{ locale.stepsList.status.completed }}</strong>
+                                {% elseif step.status == 'CANNOT_BE_DONE_YET' %}
+                                    <strong class="govuk-tag govuk-tag--purple">{{ locale.stepsList.status.cannotBeDoneYet }}</strong>
+                                {% elseif step.status == 'NO_LONGER_NEEDED' %}
+                                    <strong class="govuk-tag govuk-tag--yellow">{{ locale.stepsList.status.noLongerNeeded }}</strong>
+                                {% endif %}
```
aiden-herbert, line 103: Does Nunjuks not have `switch` statements? This seems like an ideal candidate for one if it does.

pipwilson: @aiden-herbert I don't believe so. This is the reference I use: https://mozilla.github.io/nunjucks/templating.html

aiden-herbert: I did a little delving and came across [the following](https://github.com/mozilla/nunjucks/issues/967), which suggests that it is possible and Nunjuks can do it, but it isn't under the official docs linked to demonstrate how to do it.

I thought I would ask because some say you can and others say you can't.

pipwilson: OK, so although it's not in their docs, it does look like it's in the codebase: https://github.com/mozilla/nunjucks/blob/master/nunjucks/src/parser.js#L527 - but there seems to be some concern (5 years ago!) that it could be removed in future: https://github.com/mozilla/nunjucks/issues/1283 and at least one unresolved bug report at https://github.com/mozilla/nunjucks/issues/1387. Might be a question for our esteemed leader @Thomas-Geraghty

Thomas-Geraghty: This is an interesting one. It's sat inside Nunjucks long enough that I _don't_ see it being split off, but considering the lack of documentation - I wonder if this is a smart choice for maintainability long term. I would probably just stick with `elseif` here, because if they were to remove/modify the signature for interacting with a `switch` statement, I can see it sending some future dev on a wild goose chase.

aiden-herbert: I think its silly that you can use them and have been able to for 5+ years but you cant because they don't tell you how to in the documentation or weather or not they intend to keep it in. Seems like an oversight on their part, but maintainability is key, so if they ever did decide to keep it in (or even document it), this would be a great one to make the `switch` (pun intended).

aiden-herbert (approved):

PR #384, opened 2025-02-07
Refactors assessment objects to encapsulate items better

pipwilson:
Lots of refactoring.

This PR:

1. renames "AssessmentAreas" as "FormattedAssessment" to match up with the fact that you create it through "formatAssessmentData".
2. extracts a new function `getAssessmentDetailsForArea` from `formatAssessmentData` so that it can be called for an individual area (which the per-goal pages will need later)
3. adds a new "isAssessmentComplete" flag in the FormattedAssessment object
4. rewrites the assessmentUtils.test.ts to hopefully make them more legible and maintainable


JakeGoodwin22:
Looks great! I love that we can hold a notion of assessmentIsComplete as well as the refactored structure for FormattedAssessment 🚀

JakeGoodwin22 (approved):

PR #389, opened 2025-02-12
SP2-948 - Design Feedback Changes

izzyL-DJ:
Screenshot 2025-02-12 at 10 12 18
Screenshot 2025-02-12 at 10 13 00


izzyL-DJ:
- [x] The removed content is missing a full stop: Removed on [DD Month YYYY].
- [x] In the pre-agree state ‘View steps’ link should stay as the view where the steps are visible.
- [x] More spacing between the '0 out of 1 step completed.' and the details component [10px / gov-uk spacing unit 2]
- [x] Less spacing between the details component and the 'area of need' info [20px / gov-uk spacing unit 4]

izzyL-DJ:
Currently fixing the area of need not appearing when details isn't expanded

server/views/components/summary-card/goal-summary-card.njk
```diff
@@ -107,7 +111,8 @@
                     </tbody>
               </table>
             </div>
-          </details>
+          {% if params.plan.agreementStatus == 'DRAFT' %}
+          {% endif %}</details>
```
JakeGoodwin22: What does this if statement do? Can it be removed if it is empty?

izzyL-DJ: Thanks for spotting that, was leftover code from one of the many attempts

JakeGoodwin22 (approved):
Looks good to me!

PR #393, opened 2025-02-13
SP2-1225 Add assessment info to Change Goal and Add Steps

pipwilson:


server/routes/changeGoal/ChangeGoalController.ts
```diff
@@ -34,9 +37,24 @@ export default class ChangeGoalController {
...
+      try {
+        const assessmentResponse = await req.services.assessmentService.getAssessmentByUuid(planUuid)
+        const assessmentData = assessmentResponse.sanAssessmentData
+
+        // 2. get the assessment details for the selected area of need
+        const areaConfig: AssessmentAreaConfig = areaConfigs.find(config => config.area === goal.areaOfNeed.name)
+        assessmentDetailsForArea = getAssessmentDetailsForArea(criminogenicNeedsData, areaConfig, assessmentData)
+      } catch {
+        /* swallow the error, missing data is handled in the template */
+      }
```
JakeGoodwin22: Could we avoid using 'let' by writing this block in a more functional way? Something like:

```
const assessmentDetailsForArea = await req.services.assessmentService
.getAssessmentByUuid(planUuid)
.then(assessmentResponse => {
 const assessmentData = assessmentResponse.sanAssessmentData
 const areaConfig: AssessmentAreaConfig = areaConfigs.find(config => config.area === goal.areaOfNeed.name)
 return getAssessmentDetailsForArea(criminogenicNeedsData, areaConfig, assessmentData)
})
 .catch(() => null)
```

server/routes/add-steps/AddStepsController.ts
```diff
@@ -23,6 +26,21 @@ export default class AddStepsController {
...
+      // get assessment data
+      try {
+        const assessmentResponse = await req.services.assessmentService.getAssessmentByUuid(planUuid)
+        const assessmentData = assessmentResponse.sanAssessmentData
+
+        // 2. get the assessment details for the selected area of need
+        const areaConfig: AssessmentAreaConfig = areaConfigs.find(config => config.area === goal.areaOfNeed.name)
+        assessmentDetailsForArea = getAssessmentDetailsForArea(criminogenicNeedsData, areaConfig, assessmentData)
+      } catch {
+        /* swallow the error, missing data is handled in the template */
```
aiden-herbert: Can I ask the purpose and justification of an empty catch block here? I think it may be a little concerning to rely on the template to handle the error expressly as the template should primarily display the error outcome.

server/routes/changeGoal/ChangeGoalController.ts
```diff
@@ -34,9 +37,24 @@ export default class ChangeGoalController {
...
+      // get assessment data
+      try {
+        const assessmentResponse = await req.services.assessmentService.getAssessmentByUuid(planUuid)
+        const assessmentData = assessmentResponse.sanAssessmentData
+
+        // 2. get the assessment details for the selected area of need
+        const areaConfig: AssessmentAreaConfig = areaConfigs.find(config => config.area === goal.areaOfNeed.name)
+        assessmentDetailsForArea = getAssessmentDetailsForArea(criminogenicNeedsData, areaConfig, assessmentData)
+      } catch {
+        /* swallow the error, missing data is handled in the template */
```
aiden-herbert: Same as a previous comment, can I ask the purpose and justification of an empty catch block here? I think it may be a little concerning to rely on the template to handle the error expressly as the template should primarily display the error outcome.

server/routes/add-steps/AddStepsController.ts
pipwilson: The variable that this block sets is set to null beforehand because we output a specific matching message to the user when this set of data is not available. There's no alternative fallback position if the service we call isn't available, and individual request failures are logged within the functions called here and so would be superfluous.

The catch only exists so that if the service we call isn't available then sentence plan itself doesn't crash when the error is propagated back up the stack.

server/routes/changeGoal/ChangeGoalController.ts
pipwilson: [Ditto.](https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/393#discussion_r1956077698)

server/routes/changeGoal/ChangeGoalController.ts
pipwilson: @JakeGoodwin22 resolved in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/393/commits/7f467f893e3f9244d50ac8be87f2fcc31f41b5ab

server/routes/add-steps/AddStepsController.ts
pipwilson: As part of addressing Jake's comment above I've clarified the comment in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/393/commits/7f467f893e3f9244d50ac8be87f2fcc31f41b5ab

aiden-herbert: I agree with @JakeGoodwin22 suggestion to avoid the `let` associated and appreciate the explanation :-)

server/routes/changeGoal/ChangeGoalController.ts
aiden-herbert: Same response as noted [here](https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/393#discussion_r1956096869). Thanks for looking over it and explaining :-)

aiden-herbert (approved):

PR #403, opened 2025-02-19
SP2-1388: Update error summary cards to utilise common locale instead

aiden-herbert:
Update of all hardcoded `There is a problem` titles to use the wording associated with the commonLocale. The errorSummary components have had `with context` added to those imports to allow the file access (or context) to what the commonLocale is.


pipwilson (approved):
LGTM

PR #407, opened 2025-02-26
SP2-703 Redirect to future goals tab after adding steps

pipwilson:
During the Create Goal flow and you choose to make it a goal for the future and you add steps, the "Add steps" page should redirect you back to the "future goals" tab on the Plan page. 

If the goal has a target date it should redirect you to the "goals for now" tab on the Plan page.

### DIFF
--- a/server/views/pages/add-steps.njk
+++ b/server/views/pages/add-steps.njk
@@ -77,9 +77,9 @@
     </div>
 </div>
 
-<form id="create-goal-form" method="POST">
+<form id="add-steps-form" method="POST">
     <input type="hidden" name="_csrf" value="{{ csrfToken }}"/>
-    <input type="hidden" id="_areaOfNeed" value="{{ data.areaOfNeed }}"/>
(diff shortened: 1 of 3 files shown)


server/views/pages/add-steps.njk
```diff
@@ -77,9 +77,9 @@
     </div>
 </div>
 
-<form id="create-goal-form" method="POST">
+<form id="add-steps-form" method="POST">
     <input type="hidden" name="_csrf" value="{{ csrfToken }}"/>
-    <input type="hidden" id="_areaOfNeed" value="{{ data.areaOfNeed }}"/>
```
Thomas-Geraghty, line 82: Did this have any use or was it just leftover?

pipwilson: the form had been copied/pasted from create goal - it's why the form id was wrong too

pipwilson: so no use, leftover from the copy/paste

Thomas-Geraghty (approved):

PR #409, opened 2025-02-28
SP2-1468 SP2-1469 Updates rendering to match all forms of Plan Agreement.

pipwilson:
Both of these bugs exist for the same reason - only the "Yes, I agree" answer during Plan Agreement is taken as agreement. This is incorrect - all forms of agreement count, so this change inverts the logic to check for any value except DRAFT.

I modified some of the whitespace in the area I was changing to make it match the code immediately above it so that it was easier to read - hide the whitespace when reviewing this diff, or click here: https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/409/files?diff=split&w=1

### DIFF
--- a/server/views/components/summary-card/goal-summary-card.njk
+++ b/server/views/components/summary-card/goal-summary-card.njk
@@ -56,44 +56,44 @@
         </div>
 
         {% if params.goal.steps and params.goal.steps.length > 0 %}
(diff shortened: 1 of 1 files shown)


izzyL-DJ (approved):
LGTM

PR #416, opened 2025-03-03
SP2-1439 Makes error logs clearer around plan IDs

pipwilson:
`req.session.plan?.uuid` contains the Plan Version UUID, not the Plan UUID, so remove usage of that value and replace it with Plan UUID and version number.


JakeGoodwin22 (approved):
LGTM 🚀

PR #418, opened 2025-03-04
SP2-1403: Fix token expiration check: use '>' instead of '<' to correctly validate expiry

JakeGoodwin22:
This changes the logic to only generate a new token when the current one is expired. Prior to the change, it looks like we were generating a new token every time when it was valid, but when it was invalid we were passing back the old one which was causing timeout issues to users.


pipwilson (approved):
:|

PR #421, opened 2025-03-05
SP2-1450 about page aria controls

pipwilson:
This resolution was slightly surprising.

The GDS Design System javascript looks for the spans with a class of `govuk-accordion__section-button sp-accordion-header` and inserts a button with the `aria-controls` attribute, which is generated on the fly from the root accordion ID and the current accordion item number.

Because we have four separate accordions on the page I needed to pass through a unique identifier to `renderAssessment` to insert at the root of the accordion and ensure that was also added to the content div so that it would match the generated value of `aria-controls`. This not only made the `aria-controls` and `id` attributes match up, but ensured they were unique.

Exciting!

## Before:

Screenshot 2025-03-05 at 15 29 38

## After:

Screenshot 2025-03-05 at 15 31 29


JakeGoodwin22 (approved):
Very interesting! LGTM

PR #423, opened 2025-03-06
SP2-1512 Removing todos which don't need to be done.

pipwilson:
The TODO in `plan.njk` comes from a previous design of that page, the change to having a single action on the goal summary card supersedes this TODO.

The TODO in `_needs-scores.scss` was added before I realised that the size offset was directly related to the size of the triangle pointing at the current risk score - I've added a comment to clarify this.


JakeGoodwin22 (approved):
🦤

PR #428, opened 2025-03-11
SP2-1535 Use the new endpoints in the API rather than updateGoal

pipwilson:
This PR replaces use of updateGoal with the dedicated named endpoints for the different goal actions. This matches up with the new endpoints added in https://github.com/ministryofjustice/hmpps-sentence-plan/pull/184


JakeGoodwin22 (approved):
Nice improvement!

PR #439, opened 2025-03-17
Enable ephemeral deployments

Thomas-Geraghty:


.github/workflows/ephemeral_deployment.yaml
```diff
@@ -0,0 +1,16 @@
+name: Create ephemeral deployment
+
+on:
+  pull_request:
+    types: [ synchronize, opened, reopened, labeled ]
```
alexw-moj, line 5: Do we actually want to deploy an ephemeral environment for all PRs?

alexw-moj (approved):

PR #442, opened 2025-03-19
SP2-1482 Improved Headings on About Page for Accessibility

izzyL-DJ:


server/views/components/assessment/_assessment-area.njk
```diff
@@ -1,11 +1,51 @@
 {% from "../../components/need-score/macro.njk" import needsScore %}
 {% from "components/need-score-indicator/macro.njk" import needsScoreIndicator %}
 
-<div class="govuk-accordion" data-module="govuk-accordion" id="assessment-accordion-{{ assessmentGroupId }}">
+{#{% switch assessmentGroupId %}#}
```
pipwilson: For both of these proposals you can shorten the `switch/when` and `if` statements by accessing the correct locale field by using field name indexing. i.e. you can get rid of the `switch` and `if` statements and replace them with `locale.ariaHiddenText.assessmentGroup[assessmentGroupId]`.

On a more general note about this feature, I'm not sure this is the right thing to do here. As I found when when I was doing SP2-1450 the Show/Hide link is added with JavaScript by the accordion component and really it's that link which should be getting this aria attribute, but we can't change it. Could you please check with the designers that adding visually hidden text of "Show/Hide ..." to the assessment title rather than the action link is acceptable? Make sure to let them know that any user without JavaScript will still get that visually hidden text, even though the Show or Hide action won't be on the page. Let me know if this is unclear!

server/views/pages/about.njk
```diff
@@ -122,18 +122,18 @@
             {% if data.formattedAssessmentInfo.areas.highScoring.length > 0 %}
                 {{ renderAssessment('highScoring', data.formattedAssessmentInfo.areas.highScoring, locale, data.readWrite) }}
             {% else %}
-                <p class="govuk-body">{{ locale.highScoring.sectionEmptyParagraph }}</h2>
+                <p class="govuk-body">{{ locale.highScoring.sectionEmptyParagraph }}</p>
```
pipwilson: thanks for fixing these!

server/views/components/need-score/template.njk
```diff
@@ -4,7 +4,7 @@
 
 {% set barTypeClass = 'needs-score-bar--small-' + needsScore.score + ' needs-score-bar--small-tenths' %}
 
-<div class="needs-score govuk-body">
+<div class="needs-score govuk-body" aria-hidden='true'>
```
pipwilson: double-quotes for consistency with the rest of the HTML markup please.

server/views/components/need-score-indicator/template.njk
```diff
@@ -1,5 +1,5 @@
 <div class="needs-score-label__card">
-    <p><span class="govuk-!-font-weight-bold">{{ needsScore.score }}</span>
-    out of {{ needsScore.upperBound }}</p>
+    <p><span class="govuk-!-font-weight-bold govuk-heading-s" id="needsScoreMain">{{ needsScore.score }}</span>
```
pipwilson: i haven't looked into what these IDs are doing, but they should be `class`es. An ID is used to uniquely identify an element on a web page - we have loads of needs scores on this page so we can't use an `id` attribute.

izzyL-DJ: Used this for CSS, to match the formatting from the design

pipwilson: yes, and it should still be a class, for the reasons above. https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id

server/views/components/need-score/template.njk
pipwilson: Fixed in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/442/commits/62fa9e71a7be4b8c2e64fc8707ab8debfd2485ab

assets/scss/components/_needs-scores.scss
```diff
@@ -89,6 +89,14 @@ $white-needs-score-background-colour: #ffffff;
   content: "";
 }
 
+.needsScoreMain {
+  margin-bottom: 0;
+}
+
+.needsScoreSub {
+  margin-bottom: 5px;
```
pipwilson: Here if you do `margin-bottom: govuk-spacing(1);` then sizes will auto-update if we decide to change the font size (plus it will be responsive)

integration_tests/e2e/about.cy.ts
```diff
@@ -127,7 +127,7 @@ describe('Rendering About Person for READ_WRITE user', () => {
 
     // Header assertions
     cy.get('@sectionContent')
-      .find('p.govuk-\\!-font-weight-bold')
+      .find('h4.govuk-\\!-margin-top-0')
```
pipwilson: with the previous selector when the content was inside a `p` we had to add the class to the selector to make sure we got the text that was acting as a heading. Now that they've been switched to `h4` elements do we still need the class as part of the selector?

server/views/components/need-score-indicator/template.njk
```diff
@@ -1,5 +1,5 @@
 <div class="needs-score-label__card">
-    <p><span class="govuk-!-font-weight-bold">{{ needsScore.score }}</span>
-    out of {{ needsScore.upperBound }}</p>
+    <p><span class="govuk-!-font-weight-bold govuk-heading-s needsScoreMain">{{ needsScore.score }}</span>
```
pipwilson: because you're adding your own class here you can remove the other two classes and change your CSS so it looks like this:
```css
.needsScoreMain {
  @extend .govuk-\!-font-weight-bold;
  @extend .govuk-heading-s;
  margin-bottom: 0;
}
```

integration_tests/e2e/about.cy.ts
izzyL-DJ: Nope, we do not! Shortened to jut the `h4` ✂️

pipwilson (approved):
Looking nice!

PR #444, opened 2025-03-19
SP2-1394 Handle API errors on the About page

pipwilson:
This PR adds handling for cases when either the nDelius or SAN Assessment APIs do not respond correctly.

The handling for either of those APIs being unavailable or both of them, is different, so the exceptions are caught and handled in the main flow of logic in the controller, and very specific exception displays are in the page template, rather than our usual more generic error handling.

There are a lot of file changes in this PR but that is mostly because I changed the signature of `openSentencePlan` so that it takes an `options` object parameter rather than an ever-expanding list of required parameters.


JakeGoodwin22 (approved):
🚀

PR #446, opened 2025-03-19
SP2-1504 Adds support for sorting areas which don't have a score

pipwilson:
Areas which don't have a score (which is a legitimate state) were not getting sorted by title but now they are.

### DIFF
--- a/server/utils/assessmentUtils.test.ts
+++ b/server/utils/assessmentUtils.test.ts
@@ -204,4 +204,18 @@ describe('groupAndSortAreasByRisk', () => {
       { title: 'Area A', linkedToHarm: undefined, linkedtoReoffending: undefined },
     ])
   })
+
+  it('correctly alphabetically sorts areas without scores', () => {
+    const areas: AssessmentArea[] = [
+      { title: 'Area B', linkedToHarm: 'YES', linkedtoReoffending: 'YES' } as AssessmentArea,
+      { title: 'Area A', linkedToHarm: 'YES', linkedtoReoffending: 'YES' } as AssessmentArea,
+    ]
+
+    const result = groupAndSortByRisk(areas)
+
+    expect(result).toEqual([
+      { title: 'Area A', linkedToHarm: 'YES', linkedtoReoffending: 'YES' },
+      { title: 'Area B', linkedToHarm: 'YES', linkedtoReoffending: 'YES' },
+    ])
+  })
 })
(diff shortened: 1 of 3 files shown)


JakeGoodwin22 (dismissed):
🔤

server/utils/assessmentUtils.ts
```diff
@@ -205,7 +205,8 @@ function sortByScoreAndTitle(groupedByRiskCount: Record<number, AssessmentArea[]
       groupedByRiskCount[Number(key)].sort((a, b) => {
         const distanceA = Number(a.overallScore) - a.thresholdValue
         const distanceB = Number(b.overallScore) - b.thresholdValue
-        if (distanceA === distanceB) {
+        // if the distance from thresholds are the same or if we are missing the scores, we should sort alphabetically
+        if (distanceA === distanceB || Number.isNaN(distanceA) || Number.isNaN(distanceA)) {
```
alexw-moj: Should one of these be `distanceB`?

server/utils/assessmentUtils.test.ts
```diff
@@ -204,4 +204,18 @@ describe('groupAndSortAreasByRisk', () => {
       { title: 'Area A', linkedToHarm: undefined, linkedtoReoffending: undefined },
     ])
   })
+
+  it('correctly alphabetically sorts areas without scores', () => {
```
alexw-moj, line 208: Separate tests for NaN distanceA and one for NaN distanceB, given the below?

server/utils/assessmentUtils.ts
pipwilson: ruh roh! copy/paste error! yes!

pipwilson: fixed in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/446/commits/682bc8f40eacb6a0f0c0e79e26624ff27adbc327

server/utils/assessmentUtils.test.ts
pipwilson: Interesting question. We should only be in this logic for areas which have no score (`groupAndSortByRisk(otherUnsorted)`), so both a and b should always be undefined, triggering the NaN. In principle I think that means I could actually remove the second NaN check for distanceB but I'd prefer to keep it in.

Tests for what happens if the logic in the preceding filters doesn't work or changes (i.e. tests which contain a group of areas, some with scores and some without) should be covered by some of the grouping tests above.

server/utils/assessmentUtils.ts
JakeGoodwin22: I really should've spotted this! Sorry

alexw-moj (approved):

PR #450, opened 2025-04-01
ARN-2602: Update Agreement Status after agreeing plan

JakeGoodwin22:
- Adds in the ability for a user to change the status of a plan from could not answer to either agreed or does not agree


server/views/pages/plan.njk
```diff
@@ -95,14 +95,16 @@
 
   <p class="govuk-body govuk-!-margin-top-5">
       {%- if data.plan.agreementDate !== null %}
-          {%- if data.isUpdatedAfterAgreement %}
+          {%- if data.plan.agreementStatus === 'COULD_NOT_ANSWER' and data.readWrite === true %}
+              <a class='govuk-!-display-none-print' href="/update-agree-plan">{{ locale.couldNotAgreePlanLink }}</a> {{ locale.couldNotAgreePlan }}
```
pipwilson: I think the display-none-print here is in the wrong place. It will hide only the link, not the following text.

server/@types/PlanType.ts
```diff
@@ -5,6 +5,8 @@ export enum PlanAgreementStatus {
   AGREED = 'AGREED',
   DO_NOT_AGREE = 'DO_NOT_AGREE',
   COULD_NOT_ANSWER = 'COULD_NOT_ANSWER',
+  UPDATED_AGREE = 'UPDATED_AGREE',
```
pipwilson: Need to check the values here mirror the results of the relevant comment in the API repo PR.

server/routes/plan-history/PlanHistoryController.ts
```diff
@@ -9,6 +9,13 @@ export default class PlanHistoryController {
 
     try {
       const planUuid = req.services.sessionService.getPlanUUID()
+      const planVersionNumber = req.services.sessionService.getPlanVersionNumber()
+
+      const plan =
```
pipwilson, line 14: good spot!

server/@types/PlanType.ts
JakeGoodwin22: Updated in [f5c0227](https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/450/commits/f5c02279408e5f4871a4f77049960ad18a0310e1)

server/views/pages/plan.njk
JakeGoodwin22: Updated in [f5c0227](https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/450/commits/f5c02279408e5f4871a4f77049960ad18a0310e1)

server/views/pages/plan.njk
```diff
@@ -96,10 +96,12 @@
   <p class="govuk-body govuk-!-margin-top-5">
       {%- if data.plan.agreementDate !== null %}
           {%- if data.plan.agreementStatus === 'COULD_NOT_ANSWER' and data.readWrite === true %}
-              <a class='govuk-!-display-none-print' href="/update-agree-plan">{{ locale.couldNotAgreePlanLink }}</a> {{ locale.couldNotAgreePlan }}
+              <div id='update-assessment-text' class='govuk-!-display-none-print'>
```
pipwilson: presumably this should be double quotes around the HTML attributes.

integration_tests/e2e/update-agree-plan.cy.ts
```diff
@@ -46,7 +46,7 @@ describe('Update Agree plan', () => {
     })
 
     it('Display update agree plan page correctly on load', () => {
-      cy.get('.plan-header+p').contains("Update Sam's agreement").click()
+      cy.get('#update-assessment-text').contains("Update Sam's agreement").click()
```
pipwilson: does this work? I'd have thought it should be `'#update-assessment-text > a'` or something

JakeGoodwin22: Yep this works - it searches the entire matched element, so if we really wanted to we could also get the text outside of the link to "Update Sam's agreement when you've... " (which exists within the plan-overview test)

server/views/pages/plan.njk
JakeGoodwin22: Both work, the DOM reads single quotes like double quotes but I can change for consistency. Updated in [3d298f2](https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/450/commits/3d298f2d81100c370693152b3b7c9c1bad78e9b0)

pipwilson: yeah it was just for consistency, I could have been clearer. thanks for the change.

integration_tests/e2e/update-agree-plan.cy.ts
pipwilson: but I mean the `click()` - there are two elements within the `#update-assessment-text`, the `<a>` and the inline text. Will `click()` always resolve correctly?

JakeGoodwin22: It always did in testing, but I see what you mean. I have made the change to explicitly select the `<a>` tag in [dcf013c](https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/450/commits/dcf013ca869c8599508e990e1e6c18f9477fdc9d)

server/routes/update-agree-plan/UpdateAgreePlanController.ts
```diff
@@ -0,0 +1,124 @@
...
+
+      return res.redirect(URLs.PLAN_OVERVIEW)
+    }
+
+    return next()
+  }
+
+  get = [
+    requireAccessMode(AccessMode.READ_WRITE),
+    this.validatePlanForAgreement,
```
pipwilson, line 109: I understand that this is probably based on the AgreePlanController, but because validatePlanForAgreement already fetches the Plan from the API, you could put it into the class scope so you don't need to fetch it again in the render function.

JakeGoodwin22: Sounds like a nice refactor! Updated in [b24b370](https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/450/commits/b24b370968044b0e4ffbca92a73fabef12a9cd0a)

pipwilson (approved):
Looks good to me! 🎊

PR #454, opened 2025-04-04
Remove eslint excludes

aiden-herbert:
Ensure that the ESLint is able to see the assets directory correctly and a minor update to the lint-staged check to verify correct files. This is required to allow the pre-commit hook checks to enact accordingly if you are editing files under the assets directory.

### DIFF
--- a/assets/js/autocomplete/goal-autocomplete.js
+++ b/assets/js/autocomplete/goal-autocomplete.js
@@ -15,11 +15,12 @@ async function getGoalOptionsByAreaOfNeed(areaOfNeed) {
 }
 
 async function initializeGoalInputAutocomplete() {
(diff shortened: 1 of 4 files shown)


pipwilson:
So I think we misunderstood typescript-eslint. I think now that it's essentially an extension to eslint, which is why the runtime command is still `eslint` rather than e.g. `ts-eslint`. Thus the problems in the .js files are correct (via this rule: https://eslint.org/docs/latest/rules/no-unsafe-optional-chaining), and so I've [pushed up a commit](https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/454/commits/a68281f09ade31276416c3f82f2498ec798d23dc) which reverts the lint-staged config change and fixes the javascript files. The removal of the ignore statement for eslint is kept.

pipwilson (approved):
LGTM! 😅

PR #459, opened 2025-04-24
ARN-2613 Removes the link from the gov logo and HMPPS text

pipwilson:
This markup in the HTML and SCSS mirrors that of the SAN-UI project.


JakeGoodwin22 (approved):
Looks good!

PR #460, opened 2025-04-29
ARN-2786: Fix Cypress tests

JakeGoodwin22:
Tests were calling old endpoint, causing failures. This updates them to use the remove endpoint.


pipwilson (approved):
Yep! 🆗

PR #478, opened 2025-07-01
[no-jira] VRT troubleshooting

aiden-herbert:
General troubleshooting and steps needed to run and compare snapshots generated from the CI pipeline locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,13 +55,12 @@ The test suite can be run using `make test`
 Visual Regression Testing for SP is run through the SP CI pipeline. Any failed snapshots will be uploaded as artifacts and can be accessed there.
 
 > **Note:** In order to compare failed snapshots locally, you will need to download the artifacts for each matrix container of visual regression testing.
-> These will then need to add the snapshots from the [comparison](cypress-image-diff-screenshots/comparison) folder locally into the SP project. The pipeline will also upload a JSON report for failed
-> image diffs, which will need to be incorporated into the main [cypress-image-diff-html-report](cypress-image-diff-html-report/cypress-image-diff.json). This needs to be done `before` running visual comparison locally.
+> These will then need to add the snapshots from the `comparison` and `diff` folders locally in the SP project (create these directories locally if you
+> don't have them). The pipeline will also upload a JSON report for failed image diffs, which will need to be incorporated into the main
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -55,7 +55,9 @@ The test suite can be run using `make test`
 Visual Regression Testing for SP is run through the SP CI pipeline. Any failed snapshots will be uploaded as artifacts and can be accessed there.
 
 > **Note:** In order to compare failed snapshots locally, you will need to download the artifacts for each matrix container of visual regression testing.
-> These will then need to add the snapshots from the `comparison` folder locally into the SP project. This needs to be done `before` running visual comparison locally.
+> These will then need to add the snapshots from the `comparison` and `diff` folders locally in the SP project (create these directories locally if you
```
pipwilson: should this be "need to be added" ?

cypress-image-diff-html-report/cypress-image-diff.json
pipwilson, line 1: is this file supposed to be added to git?

README.md
aiden-herbert: Yes, I shall give it a read and come up with something for grammatical correctness.

- [x] Tighten up wording

cypress-image-diff-html-report/cypress-image-diff.json
aiden-herbert: This has been added for a purpose and reason. When it comes to running the files locally, the make command needs to have something as a basis. This has been included to at least keep that running and not offer a load of faff for a dev when it comes to comparing things.

When running locally with the file as it stands, you will at least be presented with the baseline screenshots. Upon a need for comparison, you can take the generated artifact json and add those into this file OR just add that particular report file and comparison screenshots to see what has gone wrong.

Is it needed though, the directory is and the file is to ensure the make command can run as needed when investigating locally, but can either be added to with failing tests generated from the build pipeline on failure or just adding the failed build report into the same directory as the current report, so depends on choice at this time really. 

There is an iterative discussion to improve this method so that screenshots and reports are generated from the same architecture, but that is potentially more tricky. The option provided is clunky and I admit that, but we are reliant on the pipelines architecture to generate the snapshots and comparisons prior (as it is a know limitation of cypress-image-diff-js).

pipwilson (approved):
🚀

PR #490, opened 2025-09-03
Refactor containerisation

devkokov:
- Use real SAN API instead of Wiremock;
- Standardise container port numbers across ARNS projects;
- Remove Cypress container and test-up commands;
- Use Cypress GitHub Action for E2E and VRT testing;


Iana-Bezrodnykh (approved):
LGTM =)

PR #493, opened 2025-09-16
ACE-639 - CRN Based Access

Thomas-Geraghty:
- Enable access using a CRN + HMPPS auth login
- Rework sessionService to be a bit more generic for readability
- Update planService to properly return a single plan for a CRN
- Update nunjucks templates

NOTE: I would recommend reviewing this per commit rather than using the file list, because you'll end up reviewing a bunch of updated type references rather than the actual content of the PR.

### DIFF
--- a/server/services/sessionService.ts
+++ b/server/services/sessionService.ts
@@ -40,7 +40,7 @@ export default class SessionService {
     }
   }
 
-  setupPrincipleFromAuth = async (token: string) => {
+  setupPrincipalFromAuth = async (token: string) => {
     try {
       const { name, user_uuid: userUuid } = jwtDecode<JwtPayloadExtended>(token)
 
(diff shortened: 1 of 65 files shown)


server/services/sessionService.ts
```diff
@@ -14,107 +13,82 @@ export default class SessionService {
...
       }
     } catch (e) {
-      Logger.error('Failed to setup session:', e)
+      Logger.error('Failed to setup handover-based session:', e)
       throw Error(e)
     }
   }
 
-  setupAuthSession = async () => {
+  setupPrincipleFromAuth = async (token: string) => {
```
alexw-moj: Should be `Principal`?

devkokov (approved):

server/services/sessionService.ts
Thomas-Geraghty: Resolved in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/493/commits/9d8f571ecf89b315c44e3c373c1f1ffa072a08d8

PR #503, opened 2025-09-17
ACE-282 fine grained permissions

alexw-moj:
Implements more granular access controls to limit sentence plan access to the MPOP users having the case in their case load.

### DIFF
--- a/server/@types/AccessPermissions.ts
+++ b/server/@types/AccessPermissions.ts
@@ -1,6 +1,6 @@
 export type AccessPermissions = {
-  inCaseload: boolean,
-  userExcluded: boolean,
-  userRestricted: boolean,
+  inCaseload: boolean
+  userExcluded: boolean
+  userRestricted: boolean
   canAccess: boolean
 }
--- a/server/@types/AccessPermissions.ts
+++ b/server/@types/AccessPermissions.ts
@@ -1,0 +1,6 @@
+export type AccessPermissions = {
+  inCaseload: boolean,
+  userExcluded: boolean,
+  userRestricted: boolean,
+  canAccess: boolean
+}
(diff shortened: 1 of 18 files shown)


server/middleware/authorisationMiddleware.ts
```diff
@@ -15,13 +15,20 @@ export default function authorisationMiddleware(): RequestHandler {
...
     // Auth
     if (principalDetails.authType === AuthType.HMPPS_AUTH) {
-      const { authorities: roles = [] }: JwtPayloadExtended = jwtDecode(req?.user?.token)
+      const { authorities: roles = [], user_name: userName }: JwtPayloadExtended = jwtDecode(req?.user?.token)
+      const { crn } = req.services.sessionService.getSubjectDetails()
 
       if (roles.includes('ROLE_SENTENCE_PLAN')) {
-        return next()
+        const data = await req.services.sentencePlanAndDeliusService.getDataByUsernameAndCrn(userName, crn)
+
```
Thomas-Geraghty: You'll need to wrap this with a try/catch. Express doesn't handle async errors properly (or at least, in version 4 it didn't. and I wouldn't rely on version 5 - though I forget if we're on that).

server/@types/AccessPermissions.ts
Thomas-Geraghty, line 1: Not necessarily about your PR, but interesting fact - turns out you're not really meant to put bespoke types in the `@types` folder, it is meant _specifically_ for overriding external types/interfaces through namespace declarations/overrides. I can't remember why we did this initially on the project, but I've learnt this fact since.

server/middleware/authorisationMiddleware.ts
alexw-moj: Wrapped in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/503/commits/3df0ca80dc7e4bdbdc680e0d691f130a232a58ee

Thomas-Geraghty (approved):

Thomas-Geraghty (approved):

PR #516, opened 2025-09-29
ACE-738 Re-enable VRT

devkokov:
Obfuscate dynamic data on the page (such as dates) before taking a screenshot for visual regression testing


pipwilson (approved):
took me a moment to parse the regex but LGTM

PR #519, opened 2025-10-10
Nav should not show the about link if using hmpps auth

alexw-moj:


server/views/partials/layout.njk
```diff
@@ -57,9 +57,9 @@
                       text: locale.common.primaryNavigation.aboutPopText,
                       href: '/about',
                       active: data.pageId == 'about'
-                    }
-                  ]
-                }) }}
+                    } if user.authType == 'OASYS'
+          ] | reject('string')
```
pipwilson, line 61: I think something has gone wild with the indentation here - it doesn't align with the opening bracket. 

I don't understand the `reject` filter here - we use a trailing `if` on items in an array successfully in e.g. plan.njk, what is the `reject` doing/for?

alexw-moj: The if turns the item in the brackets into `""` if it's not OASYS, a bit like when we had extra tabbable to, but invisible elements in the secondary nav. The `reject` strips these out.

alexw-moj: Got Intellij to reformat the file in https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/519/commits/5155e5745cbd9b8c64549892a4f483780c9c2626

pipwilson: Formatting looks better now, thanks! Could you put a comment to the effect of the above please? Presumably this is actually a bug in the MoJ component where it should be throwing away entries which don't have all the required details to avoid this scenario. I'll file one with them.

pipwilson: https://github.com/ministryofjustice/moj-frontend/issues/1919 🤷

JakeGoodwin22: [Added comment](https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/519/commits/d7fc8c3e4bdd536f0f792128f7086efd86c8def1)

pipwilson (approved):

PR #532, opened 2025-10-27
SP2-1653: Hide 'Return to OASys' button when authenticated through hmpps auth on plan-history

JakeGoodwin22:


NicolaWard-SBG (approved):
🥇 Looks good!

PR #536, opened 2025-10-30
SP2-1651: Hide dropdown and about page links

NicolaWard-SBG:
JIRA Ticket: https://dsdmoj.atlassian.net/browse/SP2-1651

- Hide the 'View information from {name} account' dropdown when the user is not an OASys user
- Re-directs the user to the /plan overview if they try and navigate to the /about URL when the user is not an OASys user
- Refactored logic for handling the data privacy screen within Cypress tests into a reusable custom Cypress command

### DIFF
--- a/server/views/pages/change-goal.njk
+++ b/server/views/pages/change-goal.njk
@@ -7,7 +7,7 @@
 {% from "../components/error-summary/error-summary.njk" import errorSummary with context %}
 {% from "moj/components/date-picker/macro.njk" import mojDatePicker %}
 {% from "moj/components/side-navigation/macro.njk" import mojSideNavigation %}
-{% from "../components/assessment/area-assessment-summary.njk" import assessmentInformationForArea %}
+{% from "../components/assessment/area-assessment-summary.njk" import assessmentInformationForArea with context %}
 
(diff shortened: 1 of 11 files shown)


server/views/pages/change-goal.njk
```diff
@@ -7,7 +7,7 @@
 {% from "../components/error-summary/error-summary.njk" import errorSummary with context %}
 {% from "moj/components/date-picker/macro.njk" import mojDatePicker %}
 {% from "moj/components/side-navigation/macro.njk" import mojSideNavigation %}
-{% from "../components/assessment/area-assessment-summary.njk" import assessmentInformationForArea %}
+{% from "../components/assessment/area-assessment-summary.njk" import assessmentInformationForArea with context %}
```
JakeGoodwin22, line 10: What is this doing? Is it needed? Same for create-goal

NicolaWard-SBG: I thought they were needed to pass the auth context, but just tested again and they are not, will remove, thank you :)

JakeGoodwin22 (approved):
Looks good! 👍

PR #550, opened 2025-11-24
SP2-1649: Check privacy screen agreed

JakeGoodwin22:
-  Redirects to the privacy screen if accessing `/plan` without having agreed.

- Prevents user from clicking on 'Sentence Plan' on the page header to get to the plan, avoiding agreeing to the privacy statement.


NicolaWard-SBG (approved):
LGTM, had a look in the ephemeral environment and looks good.

PR #555, opened 2025-11-27
SP2-1638: Preserve previousVersion in session after Passport login to…

NicolaWard-SBG:
Part of JIRA Ticket - [SP2-1638 and SP2-1637]( https://dsdmoj.atlassian.net/browse/SP2-1637)

- Added a new route /view-previous-version/:planVersionUuid to handle viewing previous plan versions
- When this route is accessed, the previousVersion path is stored in the session.
- The user is authenticated and redirected to the historic plan version page upon successful login.
- Updated the authentication redirect logic: After login, if previousVersion exists in the session, the user is redirected there. If not, the user is redirected to the data privacy page if applicable, otherwise to the plan overview page.


JakeGoodwin22 (approved):
LGTM

PR #563, opened 2025-12-09
Sp2 1722 revert changes to hide about page

NicolaWard-SBG:
- Added back-in links that go to the 'About' page
- Added back-in links that read 'View information from [name] assessment'
- Added back-in expanders that show criminogenic needs information - ‘View information from [name] assessment’


JakeGoodwin22 (approved):
LGTM 🚀

PR #573, opened 2026-01-13
SP2-1830: Revert "Sp2 1722 revert changes to hide about page"

JakeGoodwin22:
We want to hide the about page and criminogenic info which was shown to MPoP users after making change SP2-1722


NicolaWard-SBG (approved):
I'm getting deja vu 👀 Looks good!

PR #576, opened 2026-01-27
Fix error title for change goal

Iana-Bezrodnykh:
## Problem
`change-goal` error title had "Create a goal"

Screenshot 2026-01-27 at 11 51 52

## Fix
Screenshot 2026-01-27 at 15 11 18


JakeGoodwin22 (approved):
Good catch!

PR #577, opened 2026-02-23
Dependency upgrades

JakeGoodwin22:


docker/docker-compose.dev.yml
```diff
@@ -10,10 +10,6 @@ services:
       - ../:/app:cached
       - node_modules:/app/node_modules
 
-  hmpps-auth:
-    environment:
-      JAVA_TOOL_OPTIONS: "-XX:UseSVE=0"
-
```
NicolaWard-SBG, line 16: Just a quick question on this one, why are we removing it? Could it have been put in there as a fix at somepoint?

Thomas-Geraghty: This was an issue with HMPPS Auth having issues running emulated on our ARM architectures, it's now been resolved :)

eslint.config.js
Thomas-Geraghty, line 1: I'm not sure I understand these changes? How come we changed from an `.mjs` file? What is this `minimatch` thing?

JakeGoodwin22: We had a security vulnerability in `minimatch`. After upgrading, ESLint started failing with `minimatch is not a function.` I fixed that by adding a small workaround in the config. I also changed `eslint.config` from `.mjs` to `.js` because the `.mjs` read minimatch before I could fix the not a function issue, where as in .js it let me add the workaround first

docker/docker-compose.dev.yml
NicolaWard-SBG: Thanks! :)

eslint.config.js
Thomas-Geraghty: I feel like there must be some other issue here considering I've not seen any other team have to do this, nor did we have to do it on the template

NicolaWard-SBG (approved):

eslint.config.js
JakeGoodwin22: I did look briefly to see how other teams have done it, but couldn't find anyone else who have resolved the vulnerability and have an eslint config file

